PR #1, opened 2018-04-30
Add how to build in README.md

jaebaek:
Add a simple comment to run how to build vkrunner.


bpeel:
Thanks for the patch! I made a slight tweak to remove the `./configure` because `./autogen.sh` already runs it so there’s no need to run it manually.

Issue #6, opened 2018-06-25
Support in-memory script


jaebaek:
Currently VkRunner only supports scripts in files.
In-memory scripts would be beneficial in following situations:
- On Android, File access permission is needed.
- For bug reports and reproductions, multiple file attachments would make it complicated.

I suggest decoupling file related parts from `vkrunner` and putting them in `src`.
`vr_execute()` can just receive a point to script contents buffer.

Since `vkrunner` is a library, some program can define a script contents buffer like
```c
const char *script_buffer = R"([vertex shader binary]
[...]

PR #8, opened 2018-06-26
Added string script support

jaebaek:
Previously VkRunner only supported a script in a file.
This change allows the user to define a string variable for a script.

Related issue: #6

### DIFF
--- a/vkrunner/vr-config.c
+++ b/vkrunner/vr-config.c
@@ -30,12 +30,24 @@
 #include <string.h>
 
 void
-vr_config_add_script(struct vr_config *config,
-                     const char *filename)
+vr_config_add_script_file(struct vr_config *config,
+                          const char *filename)
 {
-        struct vr_config_script *script =
-                vr_alloc(sizeof *script + strlen(filename) + 1);
-        strcpy(script->filename, filename);
+        struct vr_config_script *script;
+        script = vr_alloc(sizeof *script);
+        script->filename = vr_strdup(filename);
+        script->string = NULL;
+        vr_list_insert(config->scripts.prev, &script->link);
+}
+
+void
+vr_config_add_script_string(struct vr_config *config,
+                            const char *string)
+{
+        struct vr_config_script *script;
+        script = vr_alloc(sizeof *script);
+        script->filename = NULL;
+        script->string = vr_strdup(string);
         vr_list_insert(config->scripts.prev, &script->link);
 }
 
@@ -111,8 +123,11 @@ free_scripts(struct vr_config *config)
 {
         struct vr_config_script *script, *tmp;
 
-        vr_list_for_each_safe(script, tmp, &config->scripts, link)
+        vr_list_for_each_safe(script, tmp, &config->scripts, link) {
+                vr_free(script->filename);
+                vr_free(script->string);
                 vr_free(script);
+        }
 }
 
 static void
--- a/vkrunner/vr-script.c
+++ b/vkrunner/vr-script.c
@@ -1909,6 +1909,25 @@ read_line(FILE *f,
         return buffer->length > 0;
 }
 
+static const char *
+read_line_from_string(const char *string,
+                      struct vr_buffer *buffer)
+{
+        buffer->length = 0;
+
+        const char *end = strchr(string, '\n');
+
+        if (end == NULL) {
+                vr_buffer_append_string(buffer, string);
+        } else {
+                vr_buffer_append(buffer, string, end - string + 1);
+                vr_buffer_append_c(buffer, '\0');
+                buffer->length--;
+        }
+
+        return string + buffer->length;
+}
+
 static int
 compare_buffer_binding(const void *a,
                        const void *b)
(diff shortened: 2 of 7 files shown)


vkrunner/vr-config.c
```diff
@@ -27,18 +27,35 @@
...
-                     const char *filename)
+vr_config_add_script_file(struct vr_config *config,
+                          const char *filename)
 {
-        struct vr_config_script *script =
-                vr_alloc(sizeof *script + strlen(filename) + 1);
+        struct vr_config_script *script;
+        assert(filename && "|filename| must not be NULL");
+        script = vr_alloc(sizeof *script);
+        script->filename = vr_alloc(strlen(filename) + 1);
```
bpeel: Please use `vr_strdup()` here and in the call below.

vkrunner/vr-script.c
```diff
@@ -2027,6 +2027,25 @@ vr_script_load(const struct vr_config *config,
         return script;
 }
 
+struct vr_script *
+vr_script_load_from_string(const struct vr_config *config,
+                           const char *string)
+{
+        struct vr_script *script;
+        FILE *f = fmemopen((void *)string, strlen(string), "r");
```
bpeel: It looks like this function isn’t available on Windows. I did a quick Google and as far as I can tell it’s not available on the Android NDK either? Maybe for the sake of portability it would be better to avoid it.

Perhaps we could split `load_script_from_stream()` into `begin_load()`, `load_from_stream()`,  and `end_load()` and then for the string version we could replace `load_from_stream()` with a function that looks for lines in the string and copies them into `data.line` before calling `process_line()`. Or if you have a preference for a different way to do it, of course feel free to try it.

bpeel (changes requested):
Thanks for the patch. I have a few comments inline.

vkrunner/vr-config.c
jaebaek: Done.

vkrunner/vr-script.c
jaebaek: How about just removing handling the file stream from `vr-script.c` file and just read all file contents before starting the script parse?

For example,
1. read all file contents in `process_script()` of `vr-execute.c`.
2. `vr_script_load()` just receives the string and parses it.

bpeel: If we do that we still have write the code to split the string into lines and parse the lines one at a time. It doesn’t seem like it should be much more effort to keep the stream handling code as well. Unless there is a good reason to remove it I’d prefer if we could keep it.

jaebaek: Done.

vkrunner/vr-script.c
```diff
@@ -1862,6 +1862,41 @@ read_line(FILE *f,
         return buffer->length > 0;
 }
 
+static const char *
+read_line_from_string(const char *string,
+                      struct vr_buffer *buffer)
```
bpeel, line 1914: I think this function could be quite a bit cleaner if it scanned for the end of the line once and then added the whole line to the buffer instead of trying to add one character at a time while looking for the end of the line. It could be something like this:

```
        buffer->length = 0;

        const char *end = strchr(string, '\n');

        if (end == NULL) {
                vr_buffer_append_string(buffer, string);
        } else {
                vr_buffer_append(buffer, string, end - string + 1);
                vr_buffer_append_c(buffer, '\0');
                buffer->length--;
        }

        return string + buffer->length;
```

vkrunner/vr-config.c
```diff
@@ -27,15 +27,30 @@
...
-vr_config_add_script(struct vr_config *config,
-                     const char *filename)
+vr_config_add_script_file(struct vr_config *config,
+                          const char *filename)
 {
-        struct vr_config_script *script =
-                vr_alloc(sizeof *script + strlen(filename) + 1);
-        strcpy(script->filename, filename);
+        struct vr_config_script *script;
+        assert(filename && "|filename| must be a valid pointer");
```
bpeel: I think these asserts are not really necessary because the program would crash very quickly anyway if the assert failed. I would prefer not to have them to avoid littering the code.

vkrunner/vr-execute.c
```diff
@@ -27,6 +27,7 @@
 
 #include "vr-execute.h"
 
+#include <assert.h>
```
bpeel: This header is not used anywhere in the rest of the file.

vkrunner/vr-script.c
```diff
@@ -1923,108 +1958,181 @@ process_token_replacements(struct load_state *data)
...
-                .pipeline_keys = VR_BUFFER_STATIC_INIT,
-                .extensions = VR_BUFFER_STATIC_INIT,
-                .buffers = VR_BUFFER_STATIC_INIT,
-                .current_stage = -1,
-                .current_section = SECTION_NONE,
-                .clear_depth = 1.0f
-        };
-        bool res = true;
         int stage;
+        struct load_state *data = vr_calloc(sizeof (struct load_state));
```
bpeel: It might be nicer if load_state was allocated on the stack in the two load functions. Then load_script_begin could take a pointer to it.

vkrunner/vr-script.c
```diff
@@ -1923,108 +1958,181 @@ process_token_replacements(struct load_state *data)
...
+        assert(filename && "|filename| must be a valid pointer.");
 
-        vr_pipeline_key_init(&data.current_key);
+        data->config = config;
+        data->filename = filename;
+        data->line_num = 1;
+        data->script = vr_calloc(sizeof (struct vr_script));
+        data->current_stage = -1;
+        data->current_section = SECTION_NONE;
+        data->clear_depth = 1.0f;
```
bpeel: This is missing the initialisers for the vr_buffers. In practice I guess it will just work because the initialiser is all zeros anyway. But for the sake of consistency it would be good to call vr_buffer_init on them.

vkrunner/vr-script.c
```diff
@@ -1923,108 +1958,181 @@ process_token_replacements(struct load_state *data)
...
+
+        return res;
+}
+
+static struct vr_script *
+load_script_end(struct load_state *data)
+{
+        struct vr_script *script = data->script;
+
+        assert(data && "|data| must be a valid pointer.");
```
bpeel: If data was NULL it would have already crashed when dereferencing data->script above. I think these asserts aren’t really useful.

vkrunner/vr-script.c
```diff
@@ -1923,108 +1958,181 @@ process_token_replacements(struct load_state *data)
...
+        vr_buffer_destroy(&data->line);
+        vr_free(data);
 
-        if (res) {
-                return data.script;
-        } else {
-                vr_script_free(data.script);
-                return NULL;
-        }
+        return script;
```
bpeel: I think this bit of code could be retained here seeing as it is exactly duplicated in the two callers anyway.

bpeel (changes requested):

vkrunner/vr-script.c
jaebaek: Thank you for letting me know it.

vkrunner/vr-config.c
jaebaek: Removed.

vkrunner/vr-execute.c
jaebaek: Removed.

vkrunner/vr-script.c
jaebaek: Done.

vkrunner/vr-script.c
jaebaek: Done.

vkrunner/vr-script.c
jaebaek: Removed.

vkrunner/vr-script.c
jaebaek: The reason why I moved it to those two callers is that `res` is not accessible in this function.
Let me pass `res` to this function and retain the code.

jaebaek:
Updated based on reviews and rebased with the master branch. Please take a look.

bpeel:
Great! I’ve merged it. Thanks a lot for working on that!

Issue #5, opened 2018-06-25
Port VkRunner to Android


jaebaek:
It would be useful if VkRunner supports tests on Android.
As the first version, it may be a simple app with a simple [native activity](https://developer.android.com/ndk/guides/concepts#naa) and support only hex code shaders generated by `precompile-script.py`.

jaebaek:
@bpeel, could you please assign @jaebaek for this issue?
I cannot assign anyone. Maybe I do not have a permission to do so.

bpeel:
Ok, I have assigned you.

I just had a quick look at the Android documentation and noticed that the API includes a runtime GLSL->SPIR-V compiler. It could be nice to eventually use that. Obviously not necessary to start with though.

jaebaek:
Thank you for the suggestion. Ok let's add the support later.

PR #9, opened 2018-06-27
Ported to Android

jaebaek:
This change ports VkRunner to Android.

Related issue: #5 

One issue is that the default Android NDK compiler tools adopt flags like -Wabsolute-value, -Wint-conversion, -Wenum-conversion and they show many warning messages.

For example,

vkrunner/vr-test.c:386:38: warning: comparison between pointer and integer ('VkDescriptorSet' (aka 'unsigned long long') and 'void *')
            data->ubo_descriptor_set == NULL)
            ~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~

vkrunner/vr-test.c:657:21: warning: absolute value function 'fabsf' given an argument of type 'double' but has parameter of type 'float' which may cause truncation of value [-Wabsolute-value]
                if (fabsf(color1[p] - color2[p]) > tolerance[p])
                    ^

vkrunner/vr-test.c:104:32: warning: incompatible pointer to integer conversion assigning to 'VkBuffer' (aka 'unsigned long long') from 'void *' [-Wint-conversion]
                buffer->buffer = NULL;
                               ^ ~~~~

If we change the code based on those warning messages, it would be a major change but it is worth doing that.
@bpeel , Could you give me your opinion on it?


bpeel:
Thanks for the patch, it looks like a good start.

Where are you planning to go with this application? My expectations for an “Android port” of VkRunner would just be to make the library work on Android and perhaps make the minimal command line process run if that makes sense on Android and doesn’t imply too many changes. If you are thinking of making a fully-fledged application then maybe it would make more sense to make a separate repo for it that just imports the VkRunner library as an external dependency.

On the other hand changes to the library such as making it load the right libvulkan or using the Android API to compile GLSL->SPIR-V I think would make sense in the vkrunner repo.

As for the warnings, the ones you have given as examples look legitimate. It looks like the non-dispatchable Vulkan handles are sometimes defined as unsigned long long on some platforms (presumably Android) so comparing them with NULL is bad. I pushed a patch to make it use VK_NULL_HANDLE instead so hopefully that will fix most of them. The CTS also adds a bunch of extra compile flags which make a lot of warnings so I guess if I look at that it might fix some of the warnings on Android too.

The part at the end of the patch where you change libvulkan.so.1 to libvulkan.so looks wrong for Linux as far as I understand because that will give you the “latest ABI”. Presumably once there is a future libvulkan.so.2 then it’s going to have an incompatible ABI and VkRunner would stop working if it loaded it. If you think this is right for Android then it would need to be in an #ifdef.

I notice that the documentation for the loader says to use “hw_get_module”. Do you think that could be better than using dlopen?

jaebaek:
I want to use VkRunner to make the process of bug report and reproduction clear and efficient e.g., uploading only VkRunner script and saying that you can reproduce the bug using VkRunner.
A simple Android app that can run VkRunner is what I want.
I do not have a plan to implement an Android app more than just a command line process.
The Android app in this patch only creates a simple native activity that just runs the vkrunner library.

I updated the code related to libvulkan.so.1 as you mentioned.

I think "dlopen" is fine. Could you please point the document related to "hw_get_module"?

bpeel:
Ok, well I’m still not entirely sure whether it makes sense to have this app in VkRunner or whether it’s better in its own repo. At any rate I don’t think there’d be much benefit to merging it already until it is a bit more capable, such as being able to load a script from a file or some other way to input it to the app. I’m not really in a position to review code for an Android app, so sending your in-progress changes for review will just slow you down with no benefit anyway.

Let’s instead concentrate on the merging the library changes needed to get it to work.

For the libvulkan change, I think it would be nice if the only thing in the #ifdef was the change in name so that we don’t have to maintain two chunks of identical code. Something like this:

```C
#ifdef ANDROID
#define VULKAN_LIB "libvulkan.so"
#else
#define VULKAN_LIB "libvulkan.so.1"
#endif

        vkfn->lib_vulkan = dlopen(VULKAN_LIB, RTLD_LAZY | RTLD_GLOBAL);

        if (vkfn->lib_vulkan == NULL) {
                vr_error_message(config, "Error openining " VULKAN_LIB ": %s",
                                 dlerror());
                return false;
        }

        vkfn->vkGetInstanceProcAddr =
                dlsym(vkfn->lib_vulkan, "vkGetInstanceProcAddr");
```

The thing I read about `hw_get_module` is [here](https://vulkan.lunarg.com/doc/view/1.0.26.0/linux/LoaderAndLayerInterface.html#user-content-android). However now I see that it is talking about loading the ICD, not libvulkan.so, so it’s not relevant. Sorry for the noise.

android/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,96 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
dneto0: 2018.  :-)

android/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "string_script.inc"
+
+#include <android/log.h>
+#include <android_native_app_glue.h>
+#include <cassert>
+#include <vector>
+
+extern "C" {
+#include "config.h"
+#include <vkrunner/vkrunner.h>
```
dneto0: I think angle brackets should only be used for system includes like iostream and stdio.h.

It's not required, but I think it's good style.  You really are looking for these headers from the filesystem.

android/README.md
```diff
@@ -0,0 +1,16 @@
+VkRunner for Android
+===========
+
+This Android part code is based on Vulkan Android tutorial provided by Google.
+Since the tutorial is under Apache 2.0. license, this part of VkRunner also
+follows the license.
+
```
dneto0: Please add build instructions.

If I have guessed correctly:

1. Make sure you have the Android NDK.  (Which version works?  r15 or later I guess...)
2. Set environment variable ANDROID_NDK to point to the NDK (give an example)
3. Run ./gradlew  
4. You should have  (what APK name) where.
5. Install the APK onto the device. (What adb command?)
6. Run the APK.

Is that right?  I don't know what "gradle installDebug" does, but maybe it replaces steps 3-5?
What does "Connect to you Android device" mean?  Is it more than:
- a.  Ensure developer options are turned on
- b. connect the device via USB
- c. Verify you are connected, e.g. by doing "adb logcat"

README.md
```diff
@@ -313,3 +313,7 @@ This can by compiled using a command like the following after running
 `make install` on VkRunner:
 
     cc -o myrunner myrunner.c $(pkg-config --cflags --libs vkrunner)
+
+## Android
+
+See `android/README.md`
```
dneto0: I think you can make this a link:

     See [android/README.md](android/README.md)

?

vkrunner/vr-vk.c
```diff
@@ -71,6 +71,17 @@ vr_vk_load_libvulkan(const struct vr_config *config,
 
         vkfn->vkGetInstanceProcAddr = (void *)
                 GetProcAddress(vkfn->lib_vulkan, "vkGetInstanceProcAddr");
+#elif ANDROID
```
dneto0: Building with GCC 7.3 I get

    vr-vk.c:74:7: warning: "ANDROID" is not defined, evaluates to 0 [-Wundef]

android/app/src/main/jni/string_script.inc
```diff
@@ -0,0 +1,66 @@
+/* A sample string script. */
+
+const char *string_script = R"(
```
dneto0: Actually, I would suggest that the .inc file be *only* the raw string that you would load from a file.
That is, the main.cpp would have

```
conat char* string_script =
#include "string_script.inc"
;
```
(I would have wanted the #include inside the raw string, but then I think the preprocessor would never see the directive!)

dneto0:
@bpeel Thanks for your consideration of the PR and valuable feedback.

The app this adds is pretty much the minimal possible Android app.  Unfortunately that fact is obscured by all the boilerplate required to set up the Gradle build, app properties, etc.

@jaebaek and I discussed various flows for Android, including:

1. Bake the vkscript into the app.
   -  This makes the app less flexible, but absolutely foolproof to use.   This is the kind of flow I have used several times to report driver bugs to IHVs.  The instructions become very simple: "install the APK. Run it. Look for these messages in the Android log".  
   - Anything which adds more steps along the way to reproduce the problem greatly increases the chances that the bug is miscommunicated.

2. Have the app load the script file at runtime.
  -   That makes it more flexible and have quicker turnaround when experimenting.  I support adding that flow in the (near) future if you think it's
  -  But then you have to point the app at the filename (via command line flag? baked in path?), and you have to load the file onto the device with an "adb push" and make sure it lands in the right spot.  That adds about 3 finicky steps to a bug repro flow.  So we decided to not do this flow first.

@bpeel . What do you think?

dneto0:
> On the other hand changes to the library such as making it load the right libvulkan or using the Android API to compile GLSL->SPIR-V I think would make sense in the vkrunner repo.

FYI there is no official Android API for compiling GLSL to SPIR-V.  The recommended way of compiling shaders at runtime is to use the Shaderc library available in the NDK. See

https://developer.android.com/ndk/guides/graphics/shader-compilers 

Even though we recommend the C++ API on that page, at heart Shaderc is actually a C API that wraps around Glslang.  If the Shaderc / Glslang that ships with the NDK is too old then you can always update with latest sources from https://github.com/google/shaderc and related.

bpeel:
@dneto0 

For getting the script into the app, I think it would be kind of cool to make it into a fully-fledged GUI with a big text box. Then you could just paste the script in from wherever and press a button and it would tell you whether the test passed or not. It could also display the error messages. That way you could even have the app in the Play store and it would be very easy for people to install and run a test script.

I guess there’s no reason for the app not to be able to do all three, ie, it could also have a way to run it from the command line and load a file or use a build option to bake the script in.

I’m not really sure if my feedback is worth anything though as you both have a much better idea of how the app will be used. I’d still like to try and convince you both that it would be better off in a separate repo that just depends on VkRunner as a library. It doesn’t make sense for me to be maintaining an Android app.

android/app/src/main/jni/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+#include "string_script.inc"
+
+#include <android/log.h>
+#include <android_native_app_glue.h>
+#include <cassert>
+#include <vector>
+
+extern "C" {
```
bpeel: For what it’s worth, I changed the vkrunner headers so that they have the `extern "C"` in it themselves so they can just be directly included from C++ now.

README.md
jaebaek: Done

dneto0:
> I’d still like to try and convince you both that it would be better off in a separate repo that just depends on VkRunner as a library. It doesn’t make sense for me to be maintaining an Android app.

Point taken.  I think we can dial this back so that it's what you originally expected:  Add Android.mk support so it will be build as a library with ndk-build.

I have no ambition to make an Android app to accept the script via copy-paste.  That's way beyond my skill. :-)

jaebaek:
As we discussed, I dropped the Android app and just put the NDK build scripts.
Please take a look.

android_ndk/jni/Application.mk
```diff
@@ -0,0 +1,2 @@
+APP_STL := c++_shared
```
dneto0: I think this file should not exist.
For example, in SPIRV-Tools it only exists in the android_test subdirectory which has a tiny application showing how to build an Android executable that *uses* the SPIRV-Tools libraries.

android_ndk/jni/Android.mk
```diff
@@ -0,0 +1,65 @@
+#
```
dneto0: I think this file should be at the top level directory of vkrunner.
E.g. see https://github.com/KhronosGroup/SPIRV-Tools/blob/master/Android.mk 

It's far easier to use this way.  Particularly, when new source files are added to vkrunner, it will be easy to remember to put list them in the Android.mk file as well.

dneto0 (changes requested):

android_ndk/jni/Android.mk
```diff
@@ -0,0 +1,65 @@
...
+	$(VKRUNNER_PATH)/vr-result.c \
+	$(VKRUNNER_PATH)/vr-script.c \
+	$(VKRUNNER_PATH)/vr-subprocess.c \
+	$(VKRUNNER_PATH)/vr-temp-file.c \
+	$(VKRUNNER_PATH)/vr-test.c \
+	$(VKRUNNER_PATH)/vr-util.c \
+	$(VKRUNNER_PATH)/vr-vbo.c \
+	$(VKRUNNER_PATH)/vr-vk.c \
+	$(VKRUNNER_PATH)/vr-window.c
+
```
bpeel: It looks like the Android NDK has support for building with CMake. Would it not be possible to use this to generate the build files from the CMake files so that we don’t have to maintain two lists of source files?

https://developer.android.com/ndk/guides/cmake

android_ndk/jni/Android.mk
```diff
@@ -0,0 +1,65 @@
...
+LOCAL_C_INCLUDES := $(ROOT_PATH) $(VKRUNNER_PATH)
+LOCAL_CFLAGS	 := -Wall -Wuninitialized -Wempty-body -Wformat \
+	-Wformat-security -Winit-self -Wundef \
+	-Wvla -Wpointer-arith -Wmissing-declarations \
+	-DANDROID
```
bpeel: It looks like the Android NDK already magically defines `__ANDROID__` so we probably don’t need to add our own define.

https://stackoverflow.com/questions/8475599/is-there-an-ifdef-android-equivalent-to-ifdef-win32

README.md
```diff
@@ -313,3 +313,18 @@ This can by compiled using a command like the following after running
...
+
+VkRunner supports Android NDK build which generates the VkRunner shared library
+for Android.
+
+- Download [Android NDK](https://developer.android.com/ndk/downloads/).
+- Run the following command.
+
+```
+$ cd android_ndk/jni
+$ path/to/ndk/ndk-build
```
bpeel: These build instructions don’t seem to work without a prior build of a non-Android version because nothing generates the config.h.

bpeel (changes requested):

android_ndk/jni/Android.mk
dneto0: That CMake support is in the Gradle build system in Android Studio, and is relatively new.  The NDK is a lower level, and the Android.mk build system will work with all Android apps whether they are built with Gradle or only with the NDK.
 (Android.mk is GNU Make with extra macros and functions)

README.md
jaebaek: Thanks. Added the cmake build command to the description.

android_ndk/jni/Android.mk
jaebaek: Thanks. Changed.

android_ndk/jni/Android.mk
jaebaek: Thanks. Changed.

android_ndk/jni/Application.mk
jaebaek: This is needed to run a test on Android. I put a very simple test code for Android.

android_ndk/jni/Android.mk
jaebaek: Both CMake and Android.mk require Android NDK. Thus, building it using Android.mk does not give more required programs and it is more popular for Android build.

android/app/src/main/jni/main.cpp
jaebaek: Thanks!

jaebaek:
Updated. Please take a look.

README.md
```diff
@@ -313,3 +313,31 @@ This can by compiled using a command like the following after running
...
+for Android.
+
+- Download [Android NDK](https://developer.android.com/ndk/downloads/).
+- Run the following commands:
+
+```
+cd <vkrunner-dir>
+
+mkdir build && cd build
+cmake .. && make
```
dneto0: The generation of config.h determines whether the platform is big-endian or little-endian.  But the Android build is a cross-compile.  The answer you get from CMake is for the host platform, and so does not necessarily match the answer for the cross compile.

For the android case, you should probably use one of the native ways of getting endianness.
E.g. see https://android.googlesource.com/platform/development/+/73a5a3b/ndk/platforms/android-20/include/sys/endian.h 
It seems you can do a check such as 
#if _BYTE_ORDER == _LITTLE_ENDIAN

See also https://stackoverflow.com/a/9189572/8528814

dneto0 (changes requested):
You can't use CMake to generate the endianness for a cross-compile.

README.md
jaebaek: Changed. PTAL.

dneto0: Why do we have to run cmake and make?   

We should not have to run make because it builds things that we don't need and will throw away.

We should not have to run cmake because it is *only* needed to generate config.h.  But then even that data is overwritten by the #include of config_internal.h from vr-util.h which overwrites the preprocessor macros from config.h.   That is fragile because if include order is changed then things will break in subtle ways.

I think you should make a custom build step in Android.mk to generate the config.h from the config.h.in, and also modify config.h.in to pull in what you have in config_internal.h

dneto0 (changes requested):

dneto0:
FYI.  You can look at https://github.com/KhronosGroup/SPIRV-Tools/blob/master/Android.mk#L240 for examples of adding custom build steps for the Android case.

README.md
jaebaek: Thank you for pointing it.
I updated the code but not sure it would be fine.
Please take a look!

Android.mk
```diff
@@ -0,0 +1,66 @@
...
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+#
+
+LOCAL_PATH := $(call my-dir)
+
+$(call generate-file-dir,$(LOCAL_PATH)/config.h)
+$(LOCAL_PATH)/config.h: $(LOCAL_PATH)/config.h.android
```
dneto0: What depends on $(LOCAL_PATH)/config.h ?
Changing config.h.android should trigger rebuilding all the .c source files that depend on config.h but I don't think that dependency is captured anywhere.

config.h.android
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
```
dneto0, line 25: How did you determine these settings?   How do I know they are correct for all the platform ABIs and compiler/library combinations?

dneto0 (changes requested):

Android.mk
jaebaek: Right .. this is the part I hate Android NDK.
If you see [Android.mk document](http://android.mk/), it says "The build system handles many details for you. For example, you don't need to list header files or explicit dependencies between
  generated files in your Android.mk. The NDK build system will compute these automatically for you.".

If we use its `LOCAL_SRC_FILES`, there is no way to explicitly specify dependencies.
It also says "The LOCAL_SRC_FILES variables must contain a list of C and/or C++ source files that will be built and assembled into a module. Note that you should
**not list header and included files** here, because the build system will compute dependencies automatically for you; just list the source files
that will be passed directly to a compiler, and you should be good.".

To specify the explicit dependencies, we must use cmake instead of Android.mk.

dneto0: Ok, if it computes the header dependency on its own then I'm fine.

CMake does the same thing, by using flags -MD and -MT flags on the first compile to capture the  dependencies actually used to compile.

dneto0:
@jaebaek : I'd still like a a better answer for how you filled out config.h.android

Android.mk
jaebaek: Sorry I was wrong. I think I can put
```
config.h: config.h.android
    cp config.h.android config.h
XXX.c: config.h
```

I updated Android.mk. Please take a look.

config.h.android
jaebaek: Sorry for being late.
I updated this file.
I left only checking FFS and endian, because other options are not used in C files nor header files.

For FFS, I found that `ffs()` is declared in `libc/include/strings.h` and `libc/include/android/legacy_strings_inlines.h`
while `ffsl()` is not declared.

I check the endian based on `endian.h`.

jaebaek:
Updated. PTAL.

bpeel (approved):

jaebaek:
It seems that this change is not merged. @bpeel could you please merge it into the master branch?

bpeel:
I wasn’t sure if @dneto0 wanted to make any further comments as it is still marked as “requested changes” for him. However if you are both happy with the patch then I am happy to merge it.

jaebaek:
Sorry for this late reply. I talked with @dneto0 in offline and he thinks the PR is ok.
Thank you for merging it!

PR #11, opened 2018-08-01
examples: add a row_major example

infapi00:
When using the following line to specify matrix data:
   uniform ubo 3 mat2 16 0.0 1.0 1.0 0.0

It is assumed that the data is in the default column major mode. On
piglit, you can add a lot of info about the data, included if it is
row_major, so shader_runner would re-order the data for you.

With vkrunner you could do the same in one line with mat2 and
mat3. With mat3x4 you need to take into account the stride. As
vkrunner allows to specify the data in a more flexible way (no need to
use the shader type, that is just used to parse and push the data), it
is more pragmatic to workaround the issue by specifying the
individual columns as vec3.

### DIFF
--- a/examples/row-major.shader_test
+++ b/examples/row-major.shader_test
@@ -1,0 +1,30 @@
+[vertex shader passthrough]
+
+[fragment shader]
+#version 430
+
+layout (location = 0) out vec4 color_out;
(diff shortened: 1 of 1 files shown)


bpeel:
Thanks for the patch @infapi00.

Issue #10, opened 2018-07-13
Add image resources and samplers


jaebaek:
Some shaders use image resources and samplers for them.
To enable testing those shaders, VkRunner needs to support them.

I suggest
1. adding `image` command under `[test]` section as `ssbo` command under `[test]` section.
2. adding `sampler` command under `[test]` section.

Please give me your opinion.

bpeel:
This seems fine to start with. However there are still a lot of details to consider.

How will we set data on the image? What parameters will the `image` command take? Piglit’s shader_runner has a command to make an image with a checkerboard pattern which I think is enough for 99% of the tests.

There are quite a lot of properties on a sampler. Maybe it would make sense to have multiple commands to set them and have sensible defaults. Perhaps something semi-automatic like the thing to set properties on a pipeline could be good.

jaebaek:
I want to suggest the following syntaxes:
```
image <desc set>:<binding> <dim> <extents> <format> [values]+

sampler <desc set>:<binding> <normalized coords> <address mode> <filter mode>

sampledimage <desc set>:<binding> <normalized coords> <address mode> <filter mode> <dim> <extents> <format> [values]+
```

- `image` command defines an [sampled image](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#descriptorsets-sampledimage) with the given descriptor set, binding, dimension, extents, format, and values. Its descriptor type will be [VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkDescriptorType).
  - `<dim>` must be one of `1D, 2D, 3D, Cube, Rect, Buffer, SubpassData` as [dimensions of OpTypeImage](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Dim) describes.
  - `<extents>` must be `<x>:<y>:<z>` e.g., `640:400:1` for 640 by 400 2D image.
  - `<format>` must be the format matched to one of [VkFormat](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#features-formats-definition) e.g., `R8G8B8_UNORM` for `VK_FORMAT_R8G8B8_UNORM`.
[...]

PR #12, opened 2018-08-02
Add descriptor set option to buffer resource

jaebaek:
As a part of ideas discussed in #10 , this pull request enables setting descriptor sets for buffer resources.
The new code allows both `(ssbo|uniform ubo|probe ssbo [type]) [descriptor set]:[binding]` and `(ssbo|uniform ubo|probe ssbo [type]) [binding]`. If specified, the descriptor set must be a non-negative integer (>= 0). If it is not specified, the default descriptor set is 0 and it will be used.

### DIFF
--- a/vkrunner/vr-pipeline.c
+++ b/vkrunner/vr-pipeline.c
@@ -38,9 +38,16 @@
 #include <errno.h>
 #include <assert.h>
 #include <string.h>
+#include <limits.h>
 #ifndef WIN32
 #include <unistd.h>
 #endif
+
+struct desc_set_bindings_info {
+        VkDescriptorSetLayoutBinding *bindings;
+        unsigned n_bindings;
+        unsigned desc_set;
+};
 
 static const char *
 stage_names[VR_SCRIPT_N_STAGES] = {
@@ -663,9 +670,10 @@ create_vk_layout(struct vr_pipeline *pipeline,
         }
 
         if (pipeline->descriptor_set_layout) {
-                pipeline_layout_create_info.setLayoutCount = 1;
+                pipeline_layout_create_info.setLayoutCount =
+                        pipeline->n_desc_sets;
                 pipeline_layout_create_info.pSetLayouts =
-                        &pipeline->descriptor_set_layout;
+                        pipeline->descriptor_set_layout;
         }
 
         VkPipelineLayout layout;
@@ -682,15 +690,26 @@ create_vk_layout(struct vr_pipeline *pipeline,
         return layout;
 }
 
-static VkDescriptorSetLayout
+static bool
 create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
                                 const struct vr_script *script)
 {
         struct vr_vk *vkfn = &pipeline->window->vkfn;
         VkResult res;
+        bool ret = false;
         size_t n_buffers = script->n_buffers;
+
+        assert(n_buffers);
+
         VkDescriptorSetLayoutBinding *bindings =
                 vr_calloc(sizeof (*bindings) * n_buffers);
+        struct desc_set_bindings_info *info =
+                vr_alloc(sizeof *info * n_buffers);
+        size_t n_desc_sets = 0;
+        unsigned prev_desc_set = UINT_MAX;
+
+        unsigned n_ubo = 0;
+        unsigned n_ssbo = 0;
 
         for (unsigned i = 0; i < n_buffers; i++) {
                 const struct vr_script_buffer *buffer = script->buffers + i;
@@ -698,9 +717,11 @@ create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
                 switch (buffer->type) {
                 case VR_SCRIPT_BUFFER_TYPE_UBO:
                         descriptor_type = VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+                        ++n_ubo;
                         goto found_type;
                 case VR_SCRIPT_BUFFER_TYPE_SSBO:
                         descriptor_type = VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
+                        ++n_ssbo;
                         goto found_type;
                 }
                 vr_fatal("Unexpected buffer type");
@@ -709,30 +730,91 @@ create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
                 bindings[i].descriptorType = descriptor_type;
                 bindings[i].descriptorCount = 1;
                 bindings[i].stageFlags = pipeline->stages;
+
+                if (prev_desc_set != buffer->desc_set) {
+                        if (prev_desc_set != UINT_MAX) {
+                                info[n_desc_sets].n_bindings =
+                                        &bindings[i] -
+                                        info[n_desc_sets].bindings;
+                                ++n_desc_sets;
+                        }
+                        info[n_desc_sets].desc_set = buffer->desc_set;
+                        info[n_desc_sets].bindings = &bindings[i];
+                        prev_desc_set = buffer->desc_set;
+                }
+        }
+        info[n_desc_sets].n_bindings =
+                &bindings[n_buffers] - info[n_desc_sets].bindings;
+        ++n_desc_sets;
+
+        VkDescriptorPoolSize pool_sizes[2] = {0};
+        uint32_t n_pool_sizes = 0;
+        if (n_ubo) {
+                pool_sizes[n_pool_sizes].type =
+                        VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER;
+                pool_sizes[n_pool_sizes].descriptorCount = n_ubo;
+                n_pool_sizes++;
+        }
+        if (n_ssbo) {
+                pool_sizes[n_pool_sizes].type =
+                        VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
+                pool_sizes[n_pool_sizes].descriptorCount = n_ssbo;
+                n_pool_sizes++;
         }
 
-        VkDescriptorSetLayoutCreateInfo create_info = {
-                .sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO,
-                .bindingCount = n_buffers,
-                .pBindings = bindings
+        VkDescriptorPoolCreateInfo descriptor_pool_create_info = {
+                .sType = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO,
+                .flags = VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT,
+                .maxSets = n_desc_sets,
+                .poolSizeCount = n_pool_sizes,
+                .pPoolSizes = pool_sizes
         };
 
-        VkDescriptorSetLayout descriptor_set_layout;
-
-        res = vkfn->vkCreateDescriptorSetLayout(pipeline->window->device,
-                                                &create_info,
-                                                NULL, /* allocator */
-                                                &descriptor_set_layout);
-
-        vr_free(bindings);
-
+        res = vkfn->vkCreateDescriptorPool(pipeline->window->device,
+                                           &descriptor_pool_create_info,
+                                           NULL, /* allocator */
+                                           &pipeline->window->context->
+                                           descriptor_pool);
         if (res != VK_SUCCESS) {
                 vr_error_message(pipeline->window->config,
-                                 "Error creating descriptor set layout");
-                return VK_NULL_HANDLE;
+                                 "Error creating VkDescriptorPool");
+                goto error;
         }
 
-        return descriptor_set_layout;
+        pipeline->descriptor_set_layout =
+                vr_calloc(sizeof(VkDescriptorSetLayout) * n_desc_sets);
+        unsigned *first_sets = vr_alloc(sizeof(unsigned) * n_desc_sets);
+
+        for (unsigned i = 0; i < n_desc_sets; i++) {
+                VkDescriptorSetLayoutCreateInfo create_info = {
+                        .sType =
+                         VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO,
+                        .bindingCount = info[i].n_bindings,
+                        .pBindings = info[i].bindings
+                };
+
+                res = vkfn->vkCreateDescriptorSetLayout(
+                                pipeline->window->device,
+                                &create_info,
+                                NULL, /* allocator */
+                                &pipeline->descriptor_set_layout[i]);
+
+                if (res != VK_SUCCESS) {
+                        vr_error_message(pipeline->window->config,
+                                         "Error creating descriptor set layout");
+                        goto error;
+                }
+                first_sets[i] = info[i].desc_set;
+        }
+
+        pipeline->desc_sets = first_sets;
+        pipeline->n_desc_sets = n_desc_sets;
+        ret = true;
+
+error:
+        vr_free(bindings);
+        vr_free(info);
+        return ret;
 }
 
 struct vr_pipeline *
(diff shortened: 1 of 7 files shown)


vkrunner/vr-context.h
```diff
@@ -42,7 +42,8 @@ struct vr_context {
         VkPhysicalDeviceMemoryProperties memory_properties;
         VkPhysicalDeviceProperties device_properties;
         VkPhysicalDeviceFeatures features;
-        VkDescriptorPool descriptor_pool;
+        VkDescriptorPool *descriptor_pool;
+        unsigned n_descriptor_pool;
```
bpeel: Why does it create multiple descriptor pools? There’s nothing stopping it from allocating multiple descriptor sets from a single pool, is there?

vkrunner/vr-pipeline.c
```diff
@@ -662,9 +668,10 @@ create_vk_layout(struct vr_pipeline *pipeline,
         }
 
         if (pipeline->descriptor_set_layout) {
-                pipeline_layout_create_info.setLayoutCount = 1;
+                pipeline_layout_create_info.setLayoutCount = pipeline->
+                                window->context->n_descriptor_pool;
```
bpeel: Please try to stick to the coding style and break the line after the `=` rather than the `->`.

vkrunner/vr-pipeline.c
```diff
@@ -681,15 +688,27 @@ create_vk_layout(struct vr_pipeline *pipeline,
...
         struct vr_vk *vkfn = &pipeline->window->vkfn;
         VkResult res;
+        struct vr_buffer desc_sets;
+        struct desc_set_bindings_info *info;
+        size_t n_desc_sets;
         size_t n_buffers = script->n_buffers;
         VkDescriptorSetLayoutBinding *bindings =
                 vr_calloc(sizeof (*bindings) * n_buffers);
+        unsigned prev_desc_set = script->buffers[0].desc_set;
+
```
bpeel: Perhaps `prev_desc_set` could just start out as `UINT_MAX` and `desc_sets` could be empty. That way it doesn’t have to have this duplicated code for adding an entry in order to add the first one.

vkrunner/vr-pipeline.c
```diff
@@ -708,30 +727,91 @@ create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
...
+        for (unsigned i = 0; i < n_desc_sets; i++) {
+                res = vkfn->vkCreateDescriptorPool(pipeline->window->device,
+                                                   &descriptor_pool_create_info,
+                                                   NULL, /* allocator */
+                                                   &pipeline->window->context->
+                                                   descriptor_pool[i]);
+                if (res != VK_SUCCESS) {
+                        vr_error_message(pipeline->window->config,
+                                         "Error creating VkDescriptorPool");
+                        return;
```
bpeel: This will leak `first_sets` and `desc_sets` amongst other things on error. Maybe it could have a `goto error` thing with a cleanup phase like a lot of the other functions in VkRunner.

vkrunner/vr-pipeline.c
```diff
@@ -769,10 +849,19 @@ vr_pipeline_create(const struct vr_config *config,
         pipeline->stages = get_script_stages(script);
 
         if (script->n_buffers > 0) {
-                pipeline->descriptor_set_layout =
-                        create_vk_descriptor_set_layout(pipeline, script);
-                if (pipeline->descriptor_set_layout == VK_NULL_HANDLE)
+                create_vk_descriptor_set_layout(pipeline, script);
```
bpeel: Maybe `create_vk_descriptor_set_layout` should return `false` on error to make the error checking a little clearer. Then it wouldn’t have to have the loop below either.

vkrunner/vr-script.c
```diff
@@ -1477,15 +1492,26 @@ process_test_line(struct load_state *data)
         }
 
         if (looking_at(&p, "ssbo ")) {
-                unsigned values[2];
-                if (parse_uints(&p, values, VR_N_ELEMENTS(values), NULL)) {
+                unsigned values[3];
+                bool this_case = true;
+                const char *p_backup = p;
+                if (!(parse_uints(&p, values, 2, ":") &&
+                      parse_uints(&p, &values[2], 1, NULL))) {
```
bpeel: It might be nice to make a function to parse a binding and optional descriptor set instead of duplicating this snippet of code.

vkrunner/vr-test.c
```diff
@@ -381,31 +381,40 @@ bind_ubo_descriptor_set(struct test_data *data)
...
-                                              0, /* firstSet */
-                                              1, /* descriptorSetCount */
-                                              &data->ubo_descriptor_set,
-                                              0, /* dynamicOffsetCount */
-                                              NULL /* pDynamicOffsets */);
+        if (data->pipeline->stages & ~VK_SHADER_STAGE_COMPUTE_BIT &&
+            data->pipeline->compute_pipeline) {
+                vr_error_message(data->window->config,
+                                 "The stage is not VK_SHADER_STAGE_COMPUTE_BIT "
+                                 "but 'compute_pipeline' is not NULL");
```
bpeel: In the previous code it was possible to have a test script that has a combination of compute shaders and non-compute shaders and in that case it would bind the descriptor set to both binding points. It would be nice if you could maintain that behaviour. Perhaps the binding could be moved into a utility function with `binding_point` as a parameter.

vkrunner/vr-test.c
```diff
@@ -381,31 +381,40 @@ bind_ubo_descriptor_set(struct test_data *data)
...
+                return;
+        } else if (data->pipeline->stages & ~VK_SHADER_STAGE_COMPUTE_BIT)
+                bind_point = VK_PIPELINE_BIND_POINT_GRAPHICS;
+        else if (data->pipeline->compute_pipeline)
```
bpeel: Please try to stick to the coding style and put { } brackets around single-line if-statetements if there is another part of the if-statement that already has brackets.

bpeel (changes requested):

vkrunner/vr-context.h
jaebaek: I thought we need multiple pools for multiple sets but it was wrong.
I changed this as a single pool for all descriptor sets.

vkrunner/vr-pipeline.c
jaebaek: Done.

vkrunner/vr-pipeline.c
jaebaek: No. In the [line 731](https://github.com/Igalia/vkrunner/pull/12/files/24847f07f18723065c5f74c49ae8758c74810136#diff-e45b40913d7f4c6e7216890831e4187aR731), if the first `prev_desc_set` is `UINT_MAX`, `prev_desc_set != buffer->desc_set` is `true` even when the current `buffer->desc_set` is not the new one.

vkrunner/vr-pipeline.c
jaebaek: Done.

vkrunner/vr-pipeline.c
jaebaek: Done.

vkrunner/vr-script.c
jaebaek: Done.

vkrunner/vr-test.c
jaebaek: I did not know that a buffer can be bound both binding points. I rolled this code back and just updated `data->ubo_descriptor_set` as a pointer.

vkrunner/vr-test.c
jaebaek: Done.

jaebaek (commented):
Updated based on comments.
Thanks.

vkrunner/vr-pipeline.c
bpeel: My point was that if you remove the code to add the first entry in `desc_set` so that `desc_set` starts out empty, then when `i == 0` in the for-loop it will add the first entry using the existing code within the for-loop because `buffer->desc_set` will never be `UINT_MAX`. Ie, if `prev_desc_set` is `UINT_MAX` then it is supposed to be a new one.

bpeel:
Please try running the code with the validation layers. You can do this if you set the following environment variable:

    VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_standard_validation

It is giving a lot of new errors. If I run it on Mesa anv then it hits an assert while executing `desc_set_and_binding.shader_test`.

jaebaek:
Thanks. I fixed the problem.

dneto0:
Hi @jaebaek this needs a rebase since it now conflicts with master.

@bpeel : Aside from a rebase, are there any further issues with this PR.  We've got a few more enhancements queued up behind this one.

jaebaek:
I just rebased it.
Please take a look.

vkrunner/vr-pipeline.c
jaebaek: `prev_desc_set` exists to trigger the execution of following lines

```C
                if (prev_desc_set != buffer->desc_set) {
                        info[n_desc_sets].n_bindings =
                                &bindings[i] - info[n_desc_sets].bindings;
                        ++n_desc_sets;
                        info[n_desc_sets].desc_set = buffer->desc_set;
                        info[n_desc_sets].bindings = &bindings[i];
                        prev_desc_set = buffer->desc_set;
                }
```

When `i == 0`, it should not be triggered.
If `prev_desc_set` is initialized as `UINT_MAX`, even when `i == 0`, it triggers the above lines.

vkrunner/vr-pipeline.c
```diff
@@ -709,30 +729,105 @@ create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
...
+                pool_sizes = &pool_size_candidates[1];
+                n_pool_sizes = 1;
+        } else if (!n_ssbo) {
+                pool_sizes = &pool_size_candidates[0];
+                n_pool_sizes = 1;
+        } else {
+                pool_sizes = pool_size_candidates;
+                n_pool_sizes = 2;
+        }
+
```
bpeel: This construct seems a little odd. Maybe you could just have `pool_sizes[2]` and then do something like:

``` C
if (n_ubo) {
        pool_sizes[n_pool_sizes].type = ubo;
        pool_sizes[n_pool_sizes].descriptorCount = n_ubo;
        n_pool_sizes++;
}
/* and then the same for n_ssbos */
```

vkrunner/vr-pipeline.c
```diff
@@ -770,10 +865,17 @@ vr_pipeline_create(const struct vr_config *config,
...
-                if (pipeline->descriptor_set_layout == VK_NULL_HANDLE)
+                if (!create_vk_descriptor_set_layout(pipeline, script) ||
+                    !pipeline->descriptor_set_layout ||
+                    !pipeline->n_desc_sets) {
                         goto error;
+                } else {
+                        for (unsigned i = 0; i < pipeline->n_desc_sets; i++) {
+                                if (pipeline->descriptor_set_layout[i] ==
+                                    VK_NULL_HANDLE)
+                                        goto error;
```
bpeel: I’m still not sure why there are all these extra error checks to verify that `descriptor_set_layout` and all of its entries are not `NULL`. If any of those were the case then `create_vk_descriptor_set_layout` would have returned `false`, right? It seems better to avoid duplicating the error checking and encapsulate it in the function that handles it.

vkrunner/vr-test.c
```diff
@@ -383,30 +383,46 @@ bind_ubo_descriptor_set(struct test_data *data)
...
-                                              0, /* dynamicOffsetCount */
-                                              NULL /* pDynamicOffsets */);
+                for (unsigned i = 0; i < data->pipeline->n_desc_sets; i++) {
+                        if (data->ubo_descriptor_set[i] == VK_NULL_HANDLE) {
+                                vr_error_message(data->window->config,
+                                                 "Warning buffer descriptor set"
+                                                 " [%u] is VK_NULL_HANDLE",
+                                                 i);
+                                continue;
+                        }
```
bpeel: This check seems a little unnecessary. If it fails it means something has gone wrong internally in VkRunner, right? Maybe it should be an assert or just remove it entirely.

vkrunner/vr-test.c
```diff
@@ -383,30 +383,46 @@ bind_ubo_descriptor_set(struct test_data *data)
...
-                                              VK_PIPELINE_BIND_POINT_COMPUTE,
-                                              data->pipeline->layout,
-                                              0, /* firstSet */
-                                              1, /* descriptorSetCount */
-                                              &data->ubo_descriptor_set,
-                                              0, /* dynamicOffsetCount */
-                                              NULL /* pDynamicOffsets */);
+                for (unsigned i = 0; i < data->pipeline->n_desc_sets; i++) {
+                        if (data->ubo_descriptor_set[i] == VK_NULL_HANDLE)
+                                continue;
```
bpeel: Same here.

vkrunner/vr-pipeline.c
bpeel: I guess I’m not explaining myself very well. My suggestion was to remove these lines at the top that add the first entry in `info`:

```
        size_t n_desc_sets = 0;
        info[n_desc_sets].desc_set = prev_desc_set;
        info[n_desc_sets].bindings = bindings;
```

In that case when the `if(prev_desc_set != buffer->desc_set)` is first hit with `i==0` you *do* want it to trigger the above lines. The idea was to avoid the code duplication and only have a single place that adds entries to the `info` array in order to make it slightly more maintainable. Admittedly it is now only two lines so it’s probably not a big deal. My comment made more sense in the previous version where there was fiddly code to add an entry to the `vr_buffer`.

The downside is that you’d have to add an extra `if` to make it avoid setting `n_bindings` when adding the first entry. I guess I’ll leave it up to you which way you prefer.

vkrunner/vr-pipeline.c
jaebaek: Great. Updated it.

vkrunner/vr-pipeline.c
jaebaek: Right. `create_vk_descriptor_set_layout` checks them. Removed duplicated checks.

vkrunner/vr-test.c
jaebaek: We can remove it.
If `data->ubo_descriptor_set[i] == VK_NULL_HANDLE`, `allocate_ubo_buffers` returns `false` and `run_commands` will not be executed.
I removed these checks.

vkrunner/vr-test.c
jaebaek: Removed it.

vkrunner/vr-pipeline.c
jaebaek: Set the initial `prev_desc_set` as `UINT_MAX`.
I am not sure the new code is what you intended.
Please review it.

jaebaek (commented):
Updated it based on those comments.
Please take a look.

vkrunner/vr-pipeline.c
```diff
@@ -709,30 +730,94 @@ create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
...
+                        .bindingCount = info[i].n_bindings,
+                        .pBindings = info[i].bindings
+                };
+
+                res = vkfn->vkCreateDescriptorSetLayout(
+                                pipeline->window->device,
+                                &create_info,
+                                NULL, /* allocator */
+                                &descriptor_set_layout[i]);
+
```
bpeel: If this fails the descriptor set layouts created up to this point along with the descriptor_set_layout array will be leaked. Maybe it would be better to directly store descriptor_set_layout on the vr_pipeline when it is allocated so that vr_pipeline_free can clean it up on error. It would have to be allocated with vr_calloc and vr_pipeline_free would have to check for VK_NULL_HANDLE when freeing them.

jaebaek: I did not consider this exception. Thanks.
I  changed `VkDescriptorSetLayout *descriptor_set_layout = vr_alloc(sizeof(VkDescriptorSetLayout) * n_desc_sets);` to `pipeline->descriptor_set_layout = vr_calloc(...);` and added check for VK_NULL_HANDLE when freeing them.

PR #13, opened 2018-08-08
Add a compute shader example

jaebaek:
This example tests a compute shader written in SPIR-V code which is compiled by [clspv](https://github.com/google/clspv).


examples/compute_structure.shader_test
```diff
@@ -0,0 +1,151 @@
+# // Source code in OpenCL C:
```
bpeel: VkRunner supports a section called [comment] which it ignores. It might be nicer to put the original source code under that so that we don’t have to add # to the beginning of every line.

examples/compute_structure.shader_test
```diff
@@ -0,0 +1,151 @@
...
+VK_KHR_variable_pointers
+
+[test]
+# ssbo [bind] subdata [type] [offset] [values ...]
+ssbo 0 subdata vec4 0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0 1.0
+ssbo 1 subdata vec4 0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0 2.0
+
+compute 4 1 1
+
+# probe ssbo [tyep] [bind] [offset] == [values ...]
```
bpeel: “tyep”

bpeel (commented):
Thanks for the patch.

I wonder whether the usefulness of this example is already covered by the existing compute-shader.shader_test example. Is there anything in particular that you want to demonstrate with this one?

examples/compute_structure.shader_test
jaebaek: Done.

examples/compute_structure.shader_test
jaebaek: Done.

jaebaek:
First of all, I want to add as many examples as possible to test VkRunner itself and find bugs of VkRunner.

This example uses `VK_KHR_storage_buffer_storage_class` and `VK_KHR_variable_pointers`. I want to check some GPUs correctly support them.

The SPIR-V code is generated by OpenCL C compiler i.e., [clspv](https://github.com/google/clspv) and we also want to test it.
I have a plan to add more examples with SPIR-V code based on OpenCL C.

bpeel:
My plan for the examples directory was more along the lines of adding examples to help people write their own tests. So it is more intended as additional documentation rather than being actual useful tests. If we want to start adding tests that aren’t just documentation, maybe we could make a separate directory for them. In that case VkRunner would end up going down the lines of becoming a fully-fledged test suite rather than just a utility for test suites to use. I guess I’m undecided whether that makes sense in the long run. I was sort of hoping VkRunner would be integrated into CTS or Piglit rather than being its own test suite.

I have been adding some tests to a branch called [tests](https://github.com/Igalia/vkrunner/tree/tests). Perhaps if we make a separate directory for actual tests we could also put those there. On the other hand I would consider this only a temporary solution until we make a decision about where is the best place to have tests using VkRunner.

> This example uses `VK_KHR_storage_buffer_storage_class` and `VK_KHR_variable_pointers`. I want to check some GPUs correctly support them.

> The SPIR-V code is generated by OpenCL C compiler i.e., clspv and we also want to test it.
I have a plan to add more examples with SPIR-V code based on OpenCL C.

Seeing as VkRunner probably doesn’t do anything specific to handle code from clspv or enable those two extensions, it seems unlikely that this test would uncover bugs in VkRunner itself. They seem more like tests for the Vulkan implementation.

jaebaek:
Ok. I will keep this example in my personal repo.
Let me close this pull request.

PR #19, opened 2018-08-14
Add Vulkan validation layer

jaebaek:
This commit enables the Vulkan validation layer in source code (not by setting up the environment variable).

If this pull request is accepted, I will rebase it with [the commit for descriptor set](https://github.com/Igalia/vkrunner/pull/12).


jaebaek:
Since 3f2af290 allows VkRunner to use externally created VkDevice, I think we don't need this commit anymore.

jaebaek:
Since there is no opinion on this pull request, I will close this.

Issue #23, opened 2018-09-11
Allow error bounds for expected output


jaebaek:
Because GPUs do not provide exact precision for float values, it is hard to set the expected output.
For example, [this code](https://github.com/Igalia/vkrunner/blob/master/examples/desc_set_and_binding.shader_test#L144) has many comparisons for the reason.

Why don't we support fuzzy equality operator `~=`
and setting an epsilon value to be used across subsequent checks?
(maybe both an absolute value and relative percentage value for the epsilon with reasonable default value e.g., `0.01%`)

bpeel:
This seems like a good idea to me.

Just for info, when using the probe commands to check pixels in the framebuffer there is already a builtin tolerance which is hardcoded at 0.01. This value is taken from the default tolerance in Piglit. In Piglit’s shader_runner, there is a `tolerance` command to change this value. I wonder if we should add that command to VkRunner as well and then use the same setting for the fuzzy equality operator.

jaebaek:
That sounds good to me.

If you work for other parts, I can work for it.
Let me know that you will start it or not.

bpeel:
@jaebaek I don’t plan to work on this so please feel free to assign it to yourself.

jaebaek:
Thank you for confirming. I will send a pull request soon.

bpeel:
Fixed with https://github.com/Igalia/vkrunner/commit/a671fde90a7a6a9e860848049447a8506e7014fe

Thanks.

PR #24, opened 2018-09-12
Allow error bounds for expected output

jaebaek:
Add "tolerance" command to set up the error bound and support
fuzzy equality operator "~=". The fuzzy equality can only be used
for comparisions of float values.

EXAMPLE: examples/tolerance.shader_test
RELATED ISSUE: #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,32 @@ the largest offset.
 > probe ssbo _type_ _binding_ _offset_ _comparison_ _values_…
 
 Probes a value in the storage buffer at _binding_. The _comparison_
-can be one of `==`, `!=`, `<`, `>=`, `>` or `<=`. If the type has more
-than one component then they are compared individually until one of
-them fails the comparison.
+can be one of `==`, `!=`, `<`, `>=`, `>`, `<=` or `~=`. If the type
+has more than one component then they are compared individually until
+one of them fails the comparison. `~=` is the same with `==` but `~=`
+allows errors for `double` or `float` type numbers while `==` does
+not. Allowed errors can be set by the following `tolerance` command.
+See [examples/tolerance.shader_test](examples/tolerance.shader_test)
+for the usage of `~=`.
+
+> tolerance _tolerance0 tolerance1 tolerance2 tolerance3_
+
+Sets four tolerances i.e., allowed errors. `vecN` type values will
+use first `N` tolerances among those four. Each column of `matMxN` type
+values will also use first `N` tolerances. `float` and `double` type
+values will use only the first tolerance. Each tolerance value can be
+an `double` type real number or percentage e.g., `0.01%`. `tolerance`
+command can be also used for comparisons of pixels. See
+[examples/tolerance.shader_test]( examples/tolerance.shader_test) for
+the usage of `tolerance` command.
+
+> tolerance _tolerance0_
+
+Sets a tolerance i.e., an allowed error. If this command is set, all
+components of `vecN` and `matMxN` type values will use the same
+tolerance. Each tolerance value can be an `double` type real number or
+percentage e.g., `0.01%`. See [examples/tolerance.shader_test](
+examples/tolerance.shader_test) for the usage of `tolerance` command.
 
 > clear color _r_ _g_ _b_ _a_
 
--- a/vkrunner/vr-box.c
+++ b/vkrunner/vr-box.c
@@ -30,6 +30,7 @@
 #include <assert.h>
 
 #include "vr-util.h"
+#include "vr-tolerance.h"
 
 static const struct vr_box_type_info
 type_infos[] = {
@@ -174,6 +175,16 @@ vr_box_for_each_component(enum vr_box_type type,
         }
 }
 
+struct compare_closure {
+        enum vr_box_comparison comparison;
+        const struct vr_box *a;
+        const struct vr_box *b;
+        const struct vr_tolerance *tolerance;
+        int index;
+        int index_max;
+        bool result;
+};
+
 static bool
 compare_signed(enum vr_box_comparison comparison,
                int64_t a,
@@ -181,6 +192,7 @@ compare_signed(enum vr_box_comparison comparison,
 {
         switch (comparison) {
         case VR_BOX_COMPARISON_EQUAL:
+        case VR_BOX_COMPARISON_FUZZY_EQUAL:
                 return a == b;
         case VR_BOX_COMPARISON_NOT_EQUAL:
                 return a != b;
@@ -204,6 +216,7 @@ compare_unsigned(enum vr_box_comparison comparison,
 {
         switch (comparison) {
         case VR_BOX_COMPARISON_EQUAL:
+        case VR_BOX_COMPARISON_FUZZY_EQUAL:
                 return a == b;
         case VR_BOX_COMPARISON_NOT_EQUAL:
                 return a != b;
@@ -221,13 +234,18 @@ compare_unsigned(enum vr_box_comparison comparison,
 }
 
 static bool
-compare_double(enum vr_box_comparison comparison,
+compare_double(const struct compare_closure *data,
                double a,
                double b)
 {
-        switch (comparison) {
+        switch (data->comparison) {
         case VR_BOX_COMPARISON_EQUAL:
                 return a == b;
+        case VR_BOX_COMPARISON_FUZZY_EQUAL:
+                return vr_tolerance_equal(data->tolerance,
+                                          data->index % data->index_max,
+                                          a,
+                                          b);
         case VR_BOX_COMPARISON_NOT_EQUAL:
                 return a != b;
         case VR_BOX_COMPARISON_LESS:
@@ -244,63 +262,56 @@ compare_double(enum vr_box_comparison comparison,
 }
 
 static bool
-compare_value(enum vr_box_comparison comparison,
+compare_value(const struct compare_closure *data,
               enum vr_box_base_type type,
               const void *a,
               const void *b)
 {
         switch (type) {
         case VR_BOX_BASE_TYPE_INT:
-                return compare_signed(comparison,
+                return compare_signed(data->comparison,
                                       *(const int32_t *) a,
                                       *(const int32_t *) b);
         case VR_BOX_BASE_TYPE_UINT:
-                return compare_unsigned(comparison,
+                return compare_unsigned(data->comparison,
                                         *(const uint32_t *) a,
                                         *(const uint32_t *) b);
         case VR_BOX_BASE_TYPE_INT8:
-                return compare_signed(comparison,
+                return compare_signed(data->comparison,
                                       *(const int8_t *) a,
                                       *(const int8_t *) b);
         case VR_BOX_BASE_TYPE_UINT8:
-                return compare_unsigned(comparison,
+                return compare_unsigned(data->comparison,
                                         *(const uint8_t *) a,
                                         *(const uint8_t *) b);
         case VR_BOX_BASE_TYPE_INT16:
-                return compare_signed(comparison,
+                return compare_signed(data->comparison,
                                       *(const int16_t *) a,
                                       *(const int16_t *) b);
         case VR_BOX_BASE_TYPE_UINT16:
-                return compare_unsigned(comparison,
+                return compare_unsigned(data->comparison,
                                         *(const uint16_t *) a,
                                         *(const uint16_t *) b);
         case VR_BOX_BASE_TYPE_INT64:
-                return compare_signed(comparison,
+                return compare_signed(data->comparison,
                                       *(const int64_t *) a,
                                       *(const int64_t *) b);
         case VR_BOX_BASE_TYPE_UINT64:
-                return compare_unsigned(comparison,
+                return compare_unsigned(data->comparison,
                                         *(const uint64_t *) a,
                                         *(const uint64_t *) b);
         case VR_BOX_BASE_TYPE_FLOAT:
-                return compare_double(comparison,
+                return compare_double(data,
                                       *(const float *) a,
                                       *(const float *) b);
         case VR_BOX_BASE_TYPE_DOUBLE:
-                return compare_double(comparison,
+                return compare_double(data,
                                       *(const double *) a,
                                       *(const double *) b);
         }
 
         vr_fatal("Unexpected base type");
 }
-
-struct compare_closure {
-        enum vr_box_comparison comparison;
-        const struct vr_box *a;
-        const struct vr_box *b;
-        bool result;
-};
 
 static bool
 compare_cb(enum vr_box_base_type base_type,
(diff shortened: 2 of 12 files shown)


vkrunner/vr-box.c
```diff
@@ -182,6 +183,8 @@ compare_signed(enum vr_box_comparison comparison,
         switch (comparison) {
         case VR_BOX_COMPARISON_EQUAL:
                 return a == b;
+        case VR_BOX_COMPARISON_FUZZY_EQUAL:
+                return a == b;
```
bpeel: Perhaps this part could be combined into the case label above.

vkrunner/vr-box.c
```diff
@@ -205,6 +208,8 @@ compare_unsigned(enum vr_box_comparison comparison,
...
```
bpeel: Also here.

vkrunner/vr-box.c
```diff
@@ -223,11 +228,16 @@ compare_unsigned(enum vr_box_comparison comparison,
...
-               double b)
+               double b,
+               double tolerance,
+               bool percent)
 {
         switch (comparison) {
         case VR_BOX_COMPARISON_EQUAL:
                 return a == b;
+        case VR_BOX_COMPARISON_FUZZY_EQUAL:
+                if (percent) tolerance = tolerance / 100.0 * b;
```
bpeel: I think we might need to wrap this in `fabs` in case `b` is negative.

Please split this into two lines.

vkrunner/vr-box.c
```diff
@@ -285,11 +297,15 @@ compare_value(enum vr_box_comparison comparison,
...
                                       *(const float *) a,
-                                      *(const float *) b);
+                                      *(const float *) b,
+                                      tolerance,
+                                      percent);
         case VR_BOX_BASE_TYPE_DOUBLE:
                 return compare_double(comparison,
                                       *(const double *) a,
-                                      *(const double *) b);
+                                      *(const float *) b,
```
bpeel: This looks like a cut-and-paste mistake. I guess this should be `*(const double *)`?

vkrunner/vr-script.c
```diff
@@ -1592,6 +1604,27 @@ process_test_line(struct load_state *data)
...
         }
 
+        if (looking_at(&p, "tolerance ")) {
+                const char *p_backup = p;
+                if (parse_doubles(&p, data->tolerance, 4, NULL)) {
+                        data->tolerance_in_percent = false;
+                        return true;
+                }
+                p = p_backup;
+                if (parse_doubles(&p, data->tolerance, 4, "%")) {
```
bpeel: This is neat but I guess it will mean the final value doesn’t have to have a percent sign. Ie, this would be valid:

> tolerance 1% 1% 1% 1

I’m not sure if it really matters or not.

vkrunner/vr-script.h
```diff
@@ -36,6 +36,8 @@
 #include "vr-source.h"
 #include "vr-shader-stage.h"
 
+#define DEFAULT_TOLERANCE 0.01
+
```
bpeel: I think this should be moved into `vr-script.c` if it’s not needed elsewhere. If it is needed elsewhere it should be namespaced `VR_SCRIPT_DEFAULT_TOLERANCE`.

vkrunner/vr-test.c
```diff
@@ -663,11 +660,19 @@ static bool
...
+               bool percent,
                int n_components)
 {
-        for (int p = 0; p < n_components; ++p)
-                if (fabs(color1[p] - color2[p]) > tolerance[p])
-                        return false;
+        if (percent) {
+                for (int p = 0; p < n_components; ++p)
+                        if (fabs(color1[p] - color2[p]) >
+                            tolerance[p] / 100.0 * color2[p])
```
bpeel: I think this will need to wrapped in `fabs` too in case `color2[p]` is negative.

bpeel (commented):
Thanks for the patch.

I wonder if it’s a little confusing that the probe command only uses the first component of the tolerance when it’s comparing with an SSBO whereas it uses all of the components when comparing a pixel in the framebuffer. It is possible to use the probe command to probe a vec4 so maybe in that case it would be better to use the corresponding tolerance value for each component in order to make it more consistent. What do you think?

It might be kind of nice to make a `struct vr_tolerance` with the four tolerance values and the `bool is_percent` to avoid having to pass around the two parameters everywhere. It could even have a utility function like:

```C
bool
vr_tolerance_equal(const struct vr_tolerance *tolerance,
                   int component,
                   double value_a,
                   double value_b);
```

I think that would be useful because the code to compare against a percentage is a little bit fiddly so it would be nice to encapsulate it.

It would be great if you could also add a note in the README for the command.

It might be nice to use the fuzzy compare operator in `desc_set_and_binding.shader_test` to simplify the test a little.

jaebaek:
Thank you for the suggestion. When I wrote this code, I also thought that we need more discussions for the design. To doublecheck my understanding, it should be:

F1. `tolerance value0 value1 value2 value3` --> framebuffer, vec4 and dvec4 type
F2. `tolerance value0 value1 value2` --> vec3 and dvec3 type
F3. `tolerance value0 value1` --> vec2 and dvec2 type
F4. `tolerance value0` --> float and double type

F5. If some values are missing, use default value e.g., 0.01

What if the type is one of `mat2`, `mat2x2`, `mat2x3`, `mat3x2`, or other matrix types?

If it is `matNxM` type, using `M` values would be reasonable?
For example, we can use `tolerance value0 value1 value2` for `mat2x3` type.

bpeel:
Yes, I think that sounds like a reasonable approach if you agree too.

Yes, your explanation of matrix types sounds good to me. To clarify, I think we can treat a matNxM type as an array of N vecM’s. Eg, this command:

    probe ssbo mat2x3 0 0 ~= 1 2 3 4 5 6

would be equivalent to:

    probe ssbo vec3 0 0 ~= 1 2 3
    probe ssbo vec3 0 16 ~= 4 5 6

so it will use the tolerance values tol1, tol2, tol3, tol1, tol2, tol3 in that order. This assumes std140 and col-major layout. (I think having a way to tell the probing code and data uploading code to use a different layout could be interesting but is a separate issue).

bpeel:
> F5. If some values are missing, use default value e.g., 0.01

Hm, I think if someone specifies a single tolerance value maybe it would be more useful if it duplicated it across all of the components. It’s probably only rare cases where you want a different tolerance value for each component. I’m not sure what to do with the unspecified values if there are only two or three components. Maybe we should just disallow that? Ie, you can only specify either a single value or all four. What do you think?

jaebaek:
I agree with you. Most cases might need only a single tolerance. Let's just allow users to specify a single value or all four. I will send an updated PR soon including the description in README.md file.

vkrunner/vr-box.c
jaebaek: Done.

vkrunner/vr-box.c
jaebaek: Done

vkrunner/vr-box.c
jaebaek: "split this into two lines" is unclear to me.
I wrote `vr_tolerance_equal()` based on my understanding.
Please review it.

vkrunner/vr-box.c
jaebaek: Right. Thanks.

vkrunner/vr-script.c
jaebaek: Thanks. I updated it.

vkrunner/vr-script.h
jaebaek: Done.

vkrunner/vr-test.c
jaebaek: Done.

jaebaek (commented):
I pushed a commit based on those comments.

@bpeel 
Please let me squash those multiple commits for this pull request before merging it.
Please let me know if it looks good then I will squash it and send the squashed commit again. Then you can merge it.

README.md
```diff
@@ -120,9 +120,32 @@ the largest offset.
...
+one of them fails the comparison. `~=` is the same with `==` but `~=`
+allows errors for `double` or `float` type numbers while `==` does
+not. Allowed errors can be set by the following `tolerance` command.
+See [examples/tolerance.shader_test](examples/tolerance.shader_test)
+for the usage of `~=`.
+
+> tolerance _tolerance0 tolerance1 tolerance2 tolerance3_
+
+Sets four tolerances i.e., allowed errors. `vecN` type values will
+use first `N` tolerances among those four. Each row of `matMxN` type
```
bpeel: Shouldn’t this be “each column”?

vkrunner/vr-box.c
```diff
@@ -25,11 +25,11 @@
 
 #include "config.h"
 
-#include "vr-box.h"
-
 #include <assert.h>
 #include <math.h>
 
+#include "vr-box.h"
+#include "vr-test.h"
```
bpeel: It seems a bit messy that there is now a circular dependency between vr_box, vr_test and vr_script. Perhaps it would be cleaner to put the tolerance struct in its own header in the vr_tolerance namespace.

vkrunner/vr-box.c
```diff
@@ -331,36 +323,67 @@ compare_cb(enum vr_box_base_type base_type,
...
-                .result = true
-        };
-
-        vr_box_for_each_component(a->type, compare_cb, &data);
-
-        return data.result;
+        const struct vr_box_type_info *info = type_infos + a->type;
+        if ((info->base_type == VR_BOX_BASE_TYPE_FLOAT ||
+             info->base_type == VR_BOX_BASE_TYPE_DOUBLE) &&
+            comparison == VR_BOX_COMPARISON_FUZZY_EQUAL) {
```
bpeel: It would be nice if this could reuse the `vr_box_for_each_component` function instead of duplicating it here. Perhaps the compare closure could just have an “index” member which just counts the number of times `compare_cb` is called. Then if `vr_tolerance_equal` took a `component` argument instead of `num_components` like I originally suggested then `compare_cb` could just pass `data->index % num_rows` for the component number. That would also mean we wouldn’t have to have this `convert_float_to_double` helper.

vkrunner/vr-script.c
```diff
@@ -950,15 +951,13 @@ process_draw_rect_command(struct load_state *data,
...
-                      bool tolerance_in_percent)
+                      const struct vr_script_tolerance *tolerance)
 {
         bool relative = false;
         enum { POINT, RECT, ALL } region_type = POINT;
         int n_components;
 
-        command->probe_rect.tolerance_in_percent = tolerance_in_percent;
-        memcpy(command->probe_rect.tolerance, tolerance, 4 * sizeof(double));
+        memcpy(&command->probe_rect.tolerance, tolerance, sizeof *tolerance);
```
bpeel: This could just be

    command->probe_rect.tolerance = *tolerance;

vkrunner/vr-script.h
```diff
@@ -62,6 +60,12 @@ struct vr_script_shader {
         char source[];
 };
 
+struct vr_script_tolerance {
+        double value[4];
+        bool is_percent;
+        bool single_value;
```
bpeel: Instead of having `single_value` and then having to have special cases everywhere to handle it, why not just copy the single tolerance value into all of components when only one value is specified?

vkrunner/vr-box.c
```diff
@@ -331,36 +323,67 @@ compare_cb(enum vr_box_base_type base_type,
...
+                           data->tolerance)) {
                 data->result = false;
                 return false;
         }
 
         return true;
 }
 
+static void
+convert_float_to_double(double *dst, const float *src, int n) {
```
bpeel: Please stick to the coding style and put the `{` on its own line.

vkrunner/vr-test.c
```diff
@@ -729,15 +738,15 @@ probe_rect(struct test_data *data,
...
 
-                        if (!compare_pixels(pixel,
-                                            command->probe_rect.color,
-                                            command->probe_rect.tolerance,
-                                            percent,
-                                            n_components)) {
+                        if (!vr_test_tolerance_equal(tol,
+                                                     n_components,
+                                                     pixel,
+                                                     command->probe_rect.color)) {
```
bpeel: Please try to keep the line length to a maximum of 80 characters.

vkrunner/vr-box.c
bpeel: I was just asking not to put both the condition and the then-clause of the if-statement on the same line. Ie, it would be:

```C
if (percent)
        tolerance = tolerance / 100.0 * b;
```

But it’s irrelevant now because `vr_tolerance_equal` doesn’t have this line anymore.

README.md
jaebaek: I was confused. Thank you for correcting it.

vkrunner/vr-box.c
jaebaek: Done.

vkrunner/vr-script.h
jaebaek: Oh good idea. Thanks.

vkrunner/vr-test.c
jaebaek: Done

vkrunner/vr-script.c
jaebaek: Done

vkrunner/vr-box.c
jaebaek: Oh nice. Updated it.

vkrunner/vr-box.c
jaebaek: I couldn't think of it. Thank you for the suggestion.

jaebaek (commented):
Updated code based on comments. Please take a look.

bpeel:
I made a patch here with some further suggested changes that I thought would be easier to just change rather than describing. If you are happy with them then please feel free to squash them into your patch.

https://github.com/Igalia/vkrunner/commit/34dbd94d778e0cbf3af7186734b714155084e980

There is a further problem with the parsing of the tolerance command that I haven’t fixed in that patch. It can’t just `return is_end(p)` at the end of processing because if that fails it won’t print any error message and the script will just silently fail. I think this section of code could do with some reorganising to avoid repeating the check for negative values. Perhaps it could be moved into its own function (or even split into several functions).

jaebaek:
It looks good to me. Thanks.
After I squashed your commit with mine, I created one more commit to solve the issue you mentioned.
Feel free to update it if it is still not readable.

bpeel:
This looks good to me. If you want to squash it I will merge it.

One last thing, I noticed that the validation layers complain that the tolerance test is missing the `fragmentStoresAndAtomics` capability. Please add it.

jaebaek:
Thank you for checking it with the validation layers.
I squashed it.

Issue #25, opened 2018-09-13
Add unittests


jaebaek:
I believe that unittests will reduce unexpected bugs of VkRunner.
How do you think about adopting [googletest](https://github.com/google/googletest) for unittests?

bpeel:
Yes, it’s getting pretty urgent that we start doing some unit tests to test all the fiddly parsing corner cases. It would probably be good to look at other frameworks as well. If we had something that was C-based where we could put small tests directly into the C sources we could avoid the problem of not being able to call static functions.

jaebaek:
Sure. VkRunner uses static functions in many places. If we can test them easily, it would be nice. Please let me know your opinion after thinking of it.

bpeel:
Perhaps we could just use a very simple custom framework. If we use the test reporting stuff from CMake it doesn’t need too much code. I experimented with this in [this branch](https://github.com/Igalia/vkrunner/commits/nroberts/unit-tests).

dneto0:
Please don't make a custom unit test framework.

bpeel:
@dneto0 Can you please elaborate on your arguments against using a custom framework? Do you have another framework in mind that provides some particular features that we need?

dneto0:
Reusing code is almost always better than writing new code.

googletest is well established, powerful, and convenient.  

Being well established means new people can contribute tests and understand existing tests with low friction.  It has also sorted out lots of things like proper test isolation.

googletest is also portable, running at least on Linux (clang and GCC), macOS (clang), and Windows (using both MSVC and MinGW compilers).   Your custom code uses __attribute__((constructor)).  That's a GCC-specific syntax.

Over time your test suite is likely to grow, and you might want ever more of the power of googletest.  For example, SPIRV-Tools currently has 25410 unit tests driven via googletest, organized into 95 executables, and parallelized at the executable level by ctest.   They run in under 10s serially on my workstation, and 4s when run 20-way parallel.

You mentioned parsing corner cases as lacking in tests.  My Effcee library has a lot of parsing functionality in it.  (https://github.com/google/effcee)  It has 239 unit tests driven by googletest, in one executable, running in 0.04s.  For example the parsing of check rules starts here: https://github.com/google/effcee/blob/master/effcee/check_test.cc#L96  
I hope you can see that the expression of a unit test is simple and direct.

[...]

PR #26, opened 2018-09-21
Add googletest as unit test

jaebaek:
RELATED ISSUE: #25

### DIFF
--- a/unittest/script.cpp
+++ b/unittest/script.cpp
@@ -1,0 +1,84 @@
+// Copyright (c) 2018 Google LLC.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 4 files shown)


jaebaek:
This commit simply added googletest.
Currently many functions in [vkrunner/](vkrunner) are static functions and the unit test cannot link them. To test those static functions, we need wrappers that functions in unit tests can access.

unittest/script.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <iostream>
+#include <sstream>
+#include <stdlib.h>
+#include <string>
+
+#include "gtest/gtest.h"
+
+extern "C" {
+#include "vr-config.h"
```
samuelig, line 31:
```suggestion
#include "vr-config-private.h"
```

unittest/script.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "vr-list.h"
+#include "vr-script.h"
```
samuelig, line 33:
```suggestion
#include "vr-script-private.h"
```

samuelig:
Repository moved to https://gitlab.freedesktop.org/mesa/vkrunner. Please re-submit this pull request there.

PR #31, opened 2018-09-25
Allow probing multiple values of an SSBO with a single command

bpeel:
For example you can now probe vec2[2] with a command like:

    probe ssbo vec2 3 32 == 1 1 2 3

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,9 @@ one of them fails the comparison. `~=` is the same with `==` but `~=`
 allows errors for `double` or `float` type numbers while `==` does
 not. Allowed errors can be set by the following `tolerance` command.
 See [examples/tolerance.shader_test](examples/tolerance.shader_test)
-for the usage of `~=`.
+for the usage of `~=`. Multiple values can be listed to compare an
+array of values. In that case the buffer is assumed to have std140
+layout.
 
 > tolerance _tolerance0 tolerance1 tolerance2 tolerance3_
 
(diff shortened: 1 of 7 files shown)


jaebaek:
This looks good to me.

I have one question about the buffer layer. Based on [Vulkan spec for layout](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#interfaces-resources-layout), I think we can allow both **std140 layout** and **std430 layout**.

If it is true, adding **std430 layout** to `README.md` would be more clear for users.
If I am wrong, could please explain why **std430 layout** does not work?

bpeel:
Thanks for looking at the branch.

As far as I understand, Vulkan supports both std140 and std430 as well as completely custom layouts with arbitrary array and matrix strides for each member of the struct. std430 would work but VkRunner needs to know which one was chosen in order to calculate the offset between each member of the array. It doesn’t currently have any way for the script to inform it of the layout so VkRunner just assumes std140 throughout. This is already the case for the `ssbo subdata` command which has a similar problem.

We should probably add a way to tell VkRunner about the array and matrix strides and the choice of row/col major. However I think this is a separate [issue](https://github.com/Igalia/vkrunner/issues/32) and we should apply the same idea to the `ssbo subdata` command.

jaebaek:
Ok, let's revisit this issue when we have a use case.
This looks good to me. Thanks.

PR #35, opened 2018-10-02
Use strtod_l to parse doubles and floats

### LINKED ISSUE #34: Parsing doubles breaks with certain locales
VkRunner has the [classic parser bug](https://bugs.freedesktop.org/show_bug.cgi?id=24531) of using `strtod` and breaking under locales that don’t use a full stop (.) as the decimal separator. This doesn’t really matter for the standalone executable because it doesn’t call `setlocale`, but for the library version this could go badly wrong.

bpeel:
This fixes the problem that using strtod is locale-sensitive. This is also the approach that Mesa took to fix the same problem.

Fixes #34


bpeel:
Merged

PR #36, opened 2018-10-16
precompile-script.py: add arguments for glslangValidator and spirv-as

samuelig:
Add two new arguments to precompile-script.py, they specify where the glslangValidator and spirv-as binaries are in the system, so the user doesn't need to add them to the current path nor install them into the system.

This change is going to be used later when precompile-script.py is called by CTS in order to prebuild SPIR-V binaries.

### DIFF
--- a/precompile-script.py
+++ b/precompile-script.py
@@ -46,11 +46,12 @@ STAGE_MAP = {
 }
 
 class Converter:
-    def __init__(self, type, stage, fout):
+    def __init__(self, type, stage, fout, binary):
         self._type = type
         self._stage = stage
         self._fout = fout
         self._tempfile = tempfile.NamedTemporaryFile('w+')
+        self._binary = binary
 
     def add_line(self, line):
         self._tempfile.write(line)
@@ -60,7 +61,7 @@ class Converter:
 
         with tempfile.NamedTemporaryFile() as temp_outfile:
             if self._type == 'glsl':
-                subprocess.check_call(["glslangValidator",
+                subprocess.check_call([self._binary,
                                        "-S", self._stage,
                                        "-G",
                                        "-V",
@@ -68,7 +69,7 @@ class Converter:
                                        "-o", temp_outfile.name,
                                        self._tempfile.name])
             else:
-                subprocess.check_call(["spirv-as",
+                subprocess.check_call([self._binary,
                                        "--target-env", TARGET_ENV,
                                        "-o", temp_outfile.name,
                                        self._tempfile.name])
@@ -111,7 +112,7 @@ class Converter:
         print(file=self._fout)
 
 
-def convert_stream(fin, fout):
+def convert_stream(fin, fout, glslang, spirv_as):
     converter = None
 
     for line in fin:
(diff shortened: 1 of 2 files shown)


precompile-script.py
```diff
@@ -127,12 +128,12 @@ def convert_stream(fin, fout):
                 stage_name = section_name[:-6]
                 converter = Converter('spirv',
                                       STAGE_MAP[stage_name],
-                                      fout)
+                                      fout, spirv_as)
```
bpeel: I have a habit of putting each unrelated argument on a separate line whenever the line is too long. It might be nice to stick to that and put `spirv_as` on its own line, otherwise it looks odd.

precompile-script.py
```diff
@@ -145,28 +146,34 @@ def convert_stream(fin, fout):
         converter.finish()
 
 
+if __name__ == "__main__":
```
bpeel: If you can be bothered, it might be nice to split this change of adding the `__name == "__main__"` idiom in a separate patch or pull request.

precompile-script.py
```diff
@@ -145,28 +146,34 @@ def convert_stream(fin, fout):
...
+    parser.add_argument('-o', dest='output', metavar='OUTPUT',
+                        help='an output file or directory', required=True)
 
-parser = argparse.ArgumentParser(description='Precompile VkRunner scripts.')
-parser.add_argument('inputs', metavar='INPUT', type=str, nargs='+',
-                    help='an input file process')
-parser.add_argument('-o', dest='output', metavar='OUTPUT',
-                    help='an output file or directory', required=True)
+    parser.add_argument('-g', dest='glslang', metavar='GLSLANG',
+                        help='glslang binary path', required=False, default="./glslangValidator")
```
bpeel: Shouldn’t the default be `"glslangValidator"`? Otherwise it’s not going to look in the path for it when no `-g` option is given.

precompile-script.py
```diff
@@ -145,28 +146,34 @@ def convert_stream(fin, fout):
...
 
-args = parser.parse_args()
-output_is_directory = len(args.inputs) >= 2 or os.path.isdir(args.output)
+    parser.add_argument('-s', dest='spirv_as', metavar='SPIRV_AS',
+                        help='spirv-as binary path', required=False, default="./spirv-as")
```
bpeel: Same here.

bpeel (commented):
Thanks for the pull request. It looks good but I have a few comments below.

VkRunner itself has the environment variables `PIGLIT_GLSLANG_VALIDATOR_BINARY` and `PIGLIT_SPIRV_AS_BINARY` as a way to specify this. It might be nice to add that to this script as well. But I think for integrating into CTS it’s more explicit to have this command line option as you suggest and there’s nothing stopping us having both ways later on. We might even want to add the command line options to the VkRunner executable.

precompile-script.py
samuelig: OK

precompile-script.py
samuelig: OK

samuelig:
Pushed a commit with the respective fixes. It needs to be squashed before merging the pull request.

bpeel:
It looks good to me, thanks. Feel free to squash it and merge it whenever you want.

PR #37, opened 2018-10-18
Fix vkrunner build in CTS

samuelig:
When executing scripts/check_build_sanity.py, an error is found:

/home/siglesias/devel/vk-gl-cts/external/vkrunner/src/vkrunner/vr-pipeline.c:759:47:
   error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
        VkDescriptorPoolSize pool_sizes[2] = {0};
                                              ^
                                              {}

This patch fixes that error.

### DIFF
--- a/vkrunner/vr-pipeline.c
+++ b/vkrunner/vr-pipeline.c
@@ -756,7 +756,7 @@ create_vk_descriptor_set_layout(struct vr_pipeline *pipeline,
                 &bindings[n_buffers] - info[n_desc_sets].bindings;
         ++n_desc_sets;
 
-        VkDescriptorPoolSize pool_sizes[2] = {0};
+        VkDescriptorPoolSize pool_sizes[2];
         uint32_t n_pool_sizes = 0;
         if (n_ubo) {
                 pool_sizes[n_pool_sizes].type =


bpeel:
Thanks for the patch.

I think using an empty initialiser list is a [GCC extension](https://stackoverflow.com/questions/17589533/is-an-empty-initializer-list-valid-c-code) so I’m worried it might still have problems with some pedantic compiler options. Perhaps it would be better to just remove the initialisation? It doesn’t look like anything relies on the array being initialised to zeros.

It would be great if we could set up CI with the same build options as used in CTS.

samuelig:
> Thanks for the patch.
> 
> I think using an empty initialiser list is a [GCC extension](https://stackoverflow.com/questions/17589533/is-an-empty-initializer-list-valid-c-code) so I’m worried it might still have problems with some pedantic compiler options. Perhaps it would be better to just remove the initialisation? It doesn’t look like anything relies on the array being initialised to zeros.
> 
> It would be great if we could set up CI with the same build options as used in CTS.

Ok, I will remove the initialization then and push the commit to the repo.

PR #40, opened 2018-11-06
Replace android_test example with just building the vkrunner executable

bpeel:
Instead of the android_test example which is built as a library with a hardcoded example shader and needs to be linked to something else in order to run, we can just build the normal vkrunner executable which can be copied onto a device to run any script on the command line.


samuelig:
This seems right to me. Have you tested in an actual device? If so, feel free to merge it.

`Reviewed-by: Samuel Iglesias Gonsálvez <siglesias@igalia.com>`

bpeel:
Thanks for the review. Yes, I have been using it for internal testing on an Android device and it has already proved quite useful. I have merged it.

PR #42, opened 2018-11-15
Define VULKAN_HEADER variable to set where external vulkan headers are.

samuelig:
This way, we don't need to depend on vulkan or vulkan-headers packages. This is specially interesting for integrating VkRunner into CTS, where it generates its own headers.


bpeel:
This looks good to me, thanks. Feel free to merge it.

Issue #44, opened 2018-12-10
Error when using discontinuous descriptor sets


jaebaek:
I found that VkRunner has a bug when there are discontinuous descriptor sets.

[This](https://github.com/jaebaek/amber/blob/sparse_desc_fix/tests/cases/multiple_ssbo_with_sparse_descriptor_set_in_compute_pipeline.amber) is a script to reproduce the bug.

It works fine if descriptor sets are consecutive numbers from 0 to some positive integer `N`.
However, it does not work if descriptor sets are discontinuous e.g., `1, 3, 5`.

bpeel:
Thanks for reporting the issue. I’ve made a pull request for it in #45.

Issue #32, opened 2018-09-26
Support different buffer layouts


bpeel:
When uploading data to a UBO, SSBO or push constant, or when probing for results, VkRunner needs to know the layout of the buffer in order to calculate the offsets of the elements. The main things it needs to know are the array stride, matrix stride and the major axis of matrices. Currently VkRunner just assumes that all buffers are in std140 layout. This only matters for arrays and matrices because the layout of primitive types is only affected by their offset and that is explicitly specified in every corresponding test command anyway. Vulkan/SPIR-V supports arbitrary strides and offsets as long as they don’t conflict with certain minimums. Different strides could be specified for each member of a struct. Under GLSL as far as I know the strides are limited to a choice between std140 and std430.

There are a few ways to implement this. We could either limit the choice to std140 or std430 or we could allow a complete specification. We could make the commands per-buffer, or global or have them specific to each command that needs them.

I think my current preference would be to limit the choice to std140 and std430 and just have a global command that sets it for all subsequent commands. We could also have a command to pick the matrix axis. So something like this:

```
# Use std140 to upload these elements
layout std140
col major
ssbo 3 subdata mat3 0 1 2 3 4 5 6 7 8 9
ssbo 3 subdata float 48 10 11 12

# Use std430 for these probe commands
layout std430
row major
probe ssbo mat2 4 0 ~= 1 2 3 4
probe ssbo float 4 16 ~= 5 6 7
```

Another option might be to have a different default option for UBOs and SSBOs because that is what GLSL has. However this might be a little bit more confusing.

jaebaek:
Based on [Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#interfaces-resources-layout), Vulkan allows only `std140` and `std430` of GLSL as its layouts.
IMHO, limiting the choice to `std140` and `std430` is fine.

The example commands i.e., `layout std140` and `layout std430` look good to me, but I am not sure we need `col major` and `row major`.

infapi00:
Jumping in the discussion, as I have been involved a lot on layouts due the ARB_gl_spirv work

> Based on Vulkan spec, Vulkan allows only std140 and std430 of GLSL as its layouts.
> IMHO, limiting the choice to std140 and std430 is fine.

What that vulkan spec section points is that std140 and std430 in GLSL satisfies Vulkan memory layout rules. But Vulkan allows to extend that. For example, on that Vulkan spec quote you can read:

[...]

PR #45, opened 2018-12-12
Fix non-sequential descripter set IDs

bpeel:
The first patch fixes the bug in issue #44. The second patch is also needed to make the test case from that issue work.

Note however the test case still fails. I think it is written assuming that the test commands use the std430 layout to set data on SSBO. Perhaps that is the case for Amber. VkRunner always std140 for both SSBOs and UBOs (see issue #32) so it doesn’t work. Changing the test to use std140 makes it pass.


samuelig:
LGTM.

PR #47, opened 2018-12-19
Support configuring either std140 or std430 layouts for test commands

bpeel:
As described in issue #32, this adds commands to set the buffer layouts. The commands are:

```
push layout (std140|std430)
ssbo layout (std140|std430)
ubo layout (std140|std430)
```

Each of them sets the layouts expected for subsequent push constants, SSBOs and UBOs respectively.

There is also a patch to change the defaults for push constants and SSBOs to std430 so that it matches the defaults in GLSL. This makes writing scripts easier. However it also means that existing scripts using SSBOs might break.

Any comments would be appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,14 +85,8 @@ The same as above except that it probes the entire window.
 
 Sets a push constant at the given offset. Note that unlike Piglit, the
 offset is a byte offset into the push constant buffer rather than a
-uniform location. The type can be one of int, uint, int8_t, uint8_t,
-int16_t, uint16_t, int64_t, uint64_t, float, double, vec[234],
-dvec[234], ivec[234], uvec[234], i8vec[234], u8vec[234], i16vec[234],
-u16vec[234], i64vec[234], u64vec[234], mat[234]x[234] or
-dmat[234]x[234]. Multiple values can be specified to upload values to
-an array of that type. If matrices or arrays are specified they are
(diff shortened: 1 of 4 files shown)


dj2:
Should the commands be `uniform layout`, `ssbo layout` and `uniform ubo layout` to match what will be used to set the data later in the script?

bpeel:
@dj2 I think I prefer “(ubo|ssbo|push) layout”. If there was a “uniform layout” it wouldn’t be very intuitive and people might think it sets the layout for all uniforms, ie both push constants and UBOs.

Calling the commands “uniform” and “uniform ubo” was an attempt to make the VkRunner scripts closer to the format Piglit uses. Perhaps one day it would be better to add alternative names for these commands like this:

```
push _type_ _offset_ _values_…
ubo _binding_ subdata _type_ _offset_ _values_…
```

We could keep the old names too for compatibility.

bpeel:
I’ve pushed the patches that add the internal implementation and change the defaults to use std430 for push constants and SSBOs. There seemed to be agreement for this in the other issue. The patches that actually add the commands to change the layouts are left in this pull request in case there is further discussion.

samuelig:
Please rebase. The change LGTM.

bpeel:
Thanks, I’ve rebased and merged the branch.

PR #48, opened 2018-12-21
Fix Visual Studio 2013 build

samuelig:
There are issues with inline attribute (not supported on MSVC 2013) and offsetof() (undefined on MSVC 2013). This patch series fixes both.

Checked compilation on MSVC 2013 and GCC (GNU/Linux).

### DIFF
--- a/vkrunner/vr-buffer.h
+++ b/vkrunner/vr-buffer.h
@@ -64,7 +64,7 @@ vr_buffer_append(struct vr_buffer *buffer,
                  const void *data,
                  size_t length);
 
-static inline void
+static VR_INLINE void
 vr_buffer_append_c(struct vr_buffer *buffer,
                    char c)
 {
--- a/vkrunner/vr-feature-offsets.c
+++ b/vkrunner/vr-feature-offsets.c
@@ -24,6 +24,7 @@
  */
 
 #include "config.h"
+#include <stddef.h>
 
 #include "vr-feature-offsets.h"
 #include "vr-vk.h"
--- a/vkrunner/vr-list.h
+++ b/vkrunner/vr-list.h
@@ -26,6 +26,7 @@
 #ifndef VR_LIST_H
 #define VR_LIST_H
 
+#include <stddef.h>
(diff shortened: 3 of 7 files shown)


samuelig:
For the inline one, we can use this instead:

```C
/*
 * C99 inline keyword
 */
#ifndef inline
#  ifdef __cplusplus
     /* C++ supports inline keyword */
#  elif defined(__GNUC__)
#    define inline __inline__
#  elif defined(_MSC_VER)
#    define inline __inline
#  elif defined(__ICL)
#    define inline __inline
#  elif defined(__INTEL_COMPILER)
     /* Intel compiler supports inline keyword */
#  elif defined(__WATCOMC__) && (__WATCOMC__ >= 1100)
#    define inline __inline
#  elif (__STDC_VERSION__ >= 199901L)
     /* C99 supports inline keyword */
#  else
#    define inline
#  endif
#endif

```

Let me know what you prefer.

bpeel:
Yes, I think it would be better to continue using the C99 standard `inline` keyword on non-MSVC compilers. I don’t think it’s worth adding the ifdef’s for all of those compilers because VkRunner already heavily uses other C99 features so it would probably already have other problems. My preferred approach would be to add a VR_INLINE macro in vr-util.h that looks like this:

```C
#ifdef _MSVC_VER
#define VR_INLINE __inline
#else 
#define VR_INLINE inline
#endif 
```

Using VR_INLINE will force us to remember to include vr-util.h whenever we add another inline function.

For the second patch, it seems a bit hacky to add it to vr-util.h because nothing in vr-util uses offsetof. I think it would be better to add it to all of vr-feature-offsets.c, vr-list.h, vr-pipeline.c, vr-pipeline-key.c and vr-vk.c.

samuelig:
OK, thanks for the feedback.

samuelig:
Please ignore this.

samuelig:
Now it is fine (fixed ```#ifdef _MSC_VER```).

bpeel:
Looks good to me, thanks!

PR #49, opened 2018-12-21
Alternative uniform commands

bpeel:
Adds the following commands:

> push _type_ _offset_ _values_…

> ubo _binding_ subdata _type_ _offset_ _values_…

They are equivalent to the previous commands:

> uniform _type_ _offset_ _values_…

> uniform ubo _binding_ _type_ _offset_ _values_…

The previous commands were named that way for compatibility with Piglit but the names are a bit confusing. They are however maintained for compatibility. The `ubo subdata` command is meant to be similar to the `ssbo subdata` command.

The branch also takes the opportunity to tidy up the `process_test_line` function because it was getting very large and unwieldy.

This was discussed in issue #47.


bpeel:
I’ve pushed the patches that just tidy up the parser and don’t make any functional changes. The remaining two patches are left in this pull request in case anyone has any comments.

samuelig:
I like these two patches, they follow the naming convention in vulkan (Push constants vs uniform variables) and make it clear the usage. Fee free to push them. We probably want to update too the users of VkRunner (piglit and CTS) once we push this.

```Reviewed-by: Samuel Iglesias Gonsálvez <siglesias@igalia.com>```

bpeel:
Thanks for the review. I’ve merged it.

PR #50, opened 2018-12-24
script: fix parsing of clear commands

djdeath:
Found on Intel CI, running piglit :

./piglit run -t vulkan.shaders.fs-large-local-array-vec3  vulkan /tmp/

Signed-off-by: Lionel Landwerlin <lionel.g.landwerlin@intel.com>


bpeel:
Yikes, thanks for the fix. I have merged it. We should definitely start adding more unit tests.

PR #53, opened 2019-01-07
vkrunner: Add barriers required for copying an image to the host

### LINKED ISSUE #52: Missing barriers
I believe there are missing barriers. I have only considered the case of copying an image from the device to the host (`vkCmdCopyImageToBuffer`), but we probably need barriers for ssbos as well.

Some Android devices we tested work without the barriers, but on others nothing is rendered (there is just a black image, which is perfectly acceptable given that there are no barriers) and so the probe commands fail.

I can submit a PR soon with the barriers added.

paulthomson:
Fixes #52 (although some barriers are probably still needed for ssbos).

Note that I have removed the automatic layout transitions for the color attachment and instead opted to keep the color attachment layout as `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL`. We surround the image copy with `VkImageMemoryBarrier`s to explicitly transition to `VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL` (and back) as well as ensuring sufficient synchronization. A `VkBufferMemoryBarrier` is also added to ensure the host can safely read the data that was copied to the buffer.

This needs careful review from those who have experience with barriers.

### DIFF
--- a/vkrunner/vr-test.c
+++ b/vkrunner/vr-test.c
@@ -3,6 +3,7 @@
  *
  * Copyright (C) 2018 Neil Roberts
  * Copyright (C) 2018 Intel Coporation
+ * Copyright (C) 2019 Google LLC
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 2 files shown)


bpeel:
Thanks for the patch. I think it looks good. I too made a similar patch locally, but in the end I couldn’t convince myself whether it was actually needed or not. However I’m happy to leave the decision to someone with more experience with barriers.

It would be nice if the patch could be modified to match the coding style of the rest of code base:

* C-style /* comments */ instead of // C++
* For a function with a lot of arguments like vkCmdPipelineBarrier, it would be good to put a short comment after each one to say what the argument name is.
* Use under_scores instead of camelCase in the variable names.

paulthomson:
Thanks. I made some improvements to the style. I also changed some of the barrier parameters, which I think is better. I will try to get this checked one more time just to be sure.

chrisforbes (approved):

chrisforbes:
Absolutely necessary. Paul: I think you're oversynchronizing slightly wrt later color attachment access; but it's fairly harmless.

paulthomson:
Thanks @chrisforbes! Yeah I am happy to err on the side of caution here. In fact, we could have been much more cautious. @bpeel Let me know if you want more changes to the style. Otherwise, I am happy for this to be merged!

bpeel:
Thanks to both of you. I have merged the patch. I went ahead and made a slight change so that null pointers are represented with `NULL` instead of `0` to match the rest of the code. I hope you don’t mind.

paulthomson:
> I hope you don’t mind.

Not at all! Thanks for the merge!

PR #54, opened 2019-01-08
vkrunner: Fix typo in set_buffer_subdata

gnl21:
The offset being passed to vr_flush_memory was being taken from the wrong branch of the command structure.

### DIFF
--- a/vkrunner/vr-test.c
+++ b/vkrunner/vr-test.c
@@ -1039,7 +1039,7 @@ set_buffer_subdata(struct test_data *data,
         vr_flush_memory(data->window->context,
                         buffer->memory_type_index,
                         buffer->memory,
-                        command->set_push_constant.offset,
+                        command->set_buffer_subdata.offset,
                         command->set_buffer_subdata.size);
 
         return true;


vkrunner/vr-test.c
```diff
@@ -1039,7 +1039,7 @@ set_buffer_subdata(struct test_data *data,
         vr_flush_memory(data->window->context,
                         buffer->memory_type_index,
                         buffer->memory,
-                        command->set_push_constant.offset,
+                        command->set_buffer_subdata.offset,
                         command->set_buffer_subdata.size);
```
gnl21, line 1043: This change fixes validation problems that I'm seeing with the offset here, but I still get problems with the size. Specifically, it needs to be a multiple of the non-coherent atom size, but it typically isn't in the test. I can add some code to fix that, but is it important that this size be used here, or could it just be changed to VK_WHOLE_SIZE?

bpeel:
Thanks for the patch.

To fix the problem with it not being a multiple of nonCoherentAtomSize, I think you are right that it would probably be simpler to just flush the entire buffer. However it’s quite likely that there will be multiple small buffer updates in the script and it seems like it might potentially be inefficient to flush the entire buffer for each little update. What do you think about doing the changes in [this patch](https://github.com/Igalia/vkrunner/commit/ae0aa4ada7aa570ffd6bd98257efc2fce24079ac) instead? That makes it queue up the flushes and only do them once per buffer just before submitting the command buffer. I think this will help with fixing issue #52 as well, because that would be a point to add a memory barrier for the buffer.

The rest of the [branch](https://github.com/Igalia/vkrunner/tree/nroberts/flush-entire-buffer) adds a debug environment variable to make it always call vkFlushMappedMemoryRanges even when the memory type is coherent. The test-build.sh script would then set that variable so that we could catch any future regressions of this issue easily.

gnl21:
That patch looks good. I agree that doing multiple invalidates could be inefficient, but I wasn't sure whether it would be relevant compared to the text processing, etc. that vkrunner does. Batching up all the invalidates and doing the whole buffer at the end seems like a good solution.

bpeel:
I’ve pushed the patch that batches up the flushes. I’ve put the rest of the branch into a separate pull request (#55).

PR #55, opened 2019-01-10
Validate calls to vkFlushMappedMemoryRanges in the build test script

bpeel:
In PR #54 it was reported that the offset and size used in vkFlushMappedMemoryRanges was wrong and the validation layers were reporting it. This function is only called on platforms where the memory type is incoherent so on my Intel development machine the test build script wasn’t catching it. In order to avoid this situation regressing, this patch adds a debug environment variable to make it call vkFlushMappedMemoryRanges even when the memory is coherent. The test-build script then sets it so that the parameters to the flush command will be seen by the validation layers.

### DIFF
--- a/test-build.sh
+++ b/test-build.sh
@@ -18,6 +18,7 @@ ninja -C "$build_dir" install
 
 # Run the built executable with all of the examples and enable the
 # validation layer. Verify that nothing was written to the output.
+VKRUNNER_ALWAYS_FLUSH_MEMORY=true \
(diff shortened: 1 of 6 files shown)


samuelig:
LGTM.

```Reviewed-by: Samuel Iglesias Gonsálvez <siglesias@igalia.com>```

bpeel:
Thanks for the review. I’ve merged the branch.

Issue #46, opened 2018-12-18
Question about multiple updates for a SSBO or UBO


jaebaek:
Based on vkrunner spec, there is no limitation on updating one SSBO or UBO multiple times e.g.,
```
ssbo 0:0 subdata vec4 1 2 3 4
ssbo 0:0 subdata vec4 1 2 3 4 5 6 7 8 9 10 11 12
```

I am curious what is expected behavior for this case.

jaebaek:
In particular, I am curious the result of [this script](https://github.com/google/amber/blob/master/tests/cases/multiple_ssbo_update_with_graphics_pipeline.amber) should be a pass or failure.

dj2:
The result of the above would be an ssbo with `1 2 3 4 5 6 7 8 9 10 11 12` in it. The second call would increase the size and overwrite all the data. The ssbo is sized with the largest ssbo call that is seen.

bpeel:
Putting multiple values in an `ssbo subdata` command is intended to be able to update an array of values. So for example this line:

```
ssbo 0 subdata vec4 0     1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0 11.0 12.0
```

could be used to set the values in an SSBO declared like this:

```GLSL
layout(binding = 0, std140) buffer block {
        vec4 my_favourite_vectors[3];
}
```

Note however that VkRunner always assumes std140 layout. The default in GLSL for SSBOs is std430 so in your example VkRunner will put all the values in the wrong offsets and presumably the test won’t work. The plan is to eventually address issue #32 and have a way to tell VkRunner what buffer layout to use when updating buffers. If we do this it might be nice to make it default to std430 for SSBO updates. Whatever we do it would be great if we could try to get Amber and VkRunner to behave the same way.

bpeel:
I’ve posted a pull request (#47) to make VkRunner default to std430 for SSBOs and push constants. This makes the script mentioned [above](https://github.com/Igalia/vkrunner/issues/46#issuecomment-448358190) work.

bpeel:
I now realise I completely misunderstood your question :)

The comment from dj2 is correct, updating multiple times just overwrites any previous updates.

However note there is a subtlety here in that all drawing and compute commands are just added to the command buffer and buffer writes are implemented by just writing to the mapped memory. For example:

```
ssbo 0 subdata float 0     8.0
compute 1 1 1
ssbo 0 subdata float 0     42.0
compute 1 1 1
```

[...]

PR #56, opened 2019-01-15
Add buffer memory barriers for SSBOs at start and end of command buffer

bpeel:
As mentioned in issue #52, we need to have barriers in order to transition from accessing SSBOs from the host and in the shaders.

The first two patches change the semantics of updating SSBOs and UBOs so that they also cause the command buffer to be flushed. This is discussed in issue #46.

### DIFF
--- a/vkrunner/vr-test.c
+++ b/vkrunner/vr-test.c
@@ -190,6 +190,51 @@ begin_command_buffer(struct test_data *data)
 }
 
 static void
+add_ssbo_barriers(struct test_data *data)
+{
+        struct vr_vk *vkfn = &data->window->vkfn;
+        size_t n_barriers = 0;
+        VkBufferMemoryBarrier *barriers = NULL;
+
+        for (unsigned i = 0; i < data->script->n_buffers; i++) {
+                if (data->script->buffers[i].type != VR_SCRIPT_BUFFER_TYPE_SSBO)
+                        continue;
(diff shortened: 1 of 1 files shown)


paulthomson:
Thanks for doing this! I *believe* that you do not need the barriers for host -> device. You just need to call `vkFlushMappedMemoryRanges` (if the memory is not host coherent), then call `vkQueueSubmit` on the command buffer (in that order). All host writes flushed before the submit can be read by the device by commands in the submitted command buffer. This is due to `6.9. Host Write Ordering Guarantees` in the Vulkan spec. 

The device -> host barriers look good to me. 

I will try to get another person to verify.

bpeel:
Thanks for taking a look. If that is the case then I guess we only need the last patch and the branch doesn’t need to be dependent on making buffer updates a flush point. On the other hand it might be a good idea to do anyway to match what Amber seems to have chosen to do.

paulthomson:
I got confirmation that what I said is correct (although the code was not reviewed).

bpeel:
Thanks. I’ve force pushed the branch to contain just the last patch. I’ve moved the rest of the patches to a [separate branch](https://github.com/Igalia/vkrunner/tree/nroberts/buffer-update-flush) as it’s now unrelated.

paulthomson:
Looks good to me!

PR #60, opened 2019-02-19
Add float16 types

bpeel:
Adds float16 types for setting values in uniforms and SSBOs as well as for probing SSBOs.


samuelig:
LGTM

PR #61, opened 2019-03-27
script: Allow specifying the required API version

hikiko:
Allows the user to set the required version in the require section like
that:

```
[require]
vulkan <version>
```

Fixes: Issue 29

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,14 +85,8 @@ The same as above except that it probes the entire window.
 
 Sets a push constant at the given offset. Note that unlike Piglit, the
 offset is a byte offset into the push constant buffer rather than a
-uniform location. The type can be one of int, uint, int8_t, uint8_t,
-int16_t, uint16_t, int64_t, uint64_t, float16_t, float, double,
-f16vec[234], vec[234], dvec[234], ivec[234], uvec[234], i8vec[234],
-u8vec[234], i16vec[234], u16vec[234], i64vec[234], u64vec[234],
-mat[234]x[234] or dmat[234]x[234]. Multiple values can be specified to
-upload values to an array of that type. If matrices or arrays are
-specified they are assumed to have a stride according to layout
-specified with the `push layout` command.
+uniform location. For a description of how the arguments work see
+“Setting buffer subdata” below.
 
 > (ubo|ssbo) _binding_ subdata _type_ _offset_ _values_…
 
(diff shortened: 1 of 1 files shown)


bpeel:
Thanks for the patch. It looks like a good start.

I think it would be good to put the version requirement directly into the `vr_requirements` struct instead of having to pass another argument to `vr_context_new`. That struct is supposed to represent everything that is required by the script to create the context.

We probably need to do something when creating the device in order to request the corresponding version. This might just need a small tweak to `vk_init_device`.

It would be nice to pass the version to glslang via the --target-env option. This might also involve tweaking the precompile script.

hikiko:
Ok, I am going to work on the changes, thank you!

bpeel:
Thanks for updating the patches. I have some feedback below:

### First two patches

I think it would be better to squash these two patches together.

We should check the API version in vr_requirements_check instead of doing it after creating the device. That way it can use the information to pick the right physical device.

We should probably also check the API version *before* creating the instance. This can be done as described in the following comment in the API docs:

> Because Vulkan 1.0 implementations may fail with VK_ERROR_INCOMPATIBLE_DRIVER, applications should determine the version of Vulkan available before calling vkCreateInstance. If the vkGetInstanceProcAddr returns NULL for vkEnumerateInstanceVersion, it is a Vulkan 1.0 implementation. Otherwise, the application can call vkEnumerateInstanceVersion to determine the version of Vulkan.

So if the version is > 1.0 then we should check if the `vkEnumerateInstanceVersion` function is available. We don’t actually need to call it because apparently as long as the instance version is at least 1.1 then we can still use devices that support > 1.1.

We need to add something to handle the version in `vr_requirements_equal` and `vr_requirements_copy`.

We should initialise the version to something in vr_requirements_new. Otherwise it will try to create a Vulkan instance with API version 0.0.0 when the version is not specified in the script.

It would be nice if you could add a note about the vulkan line in the README.

### Changed the precompile-script target_env option

The `.` in the VERSION_RE regular expression should be `\.`, otherwise it matches any character. Instead of `[ \t]` it might be simpler to just use `\s`.

We can remove the `def add_version` method as it’s not called anywhere.

Instead of doing `re.sub(…, "\1")` to get the first match, you can just call `vmd.group(1)`.

The `elif section_name.endswith('require')` should just be `section_name == 'require'`.

I think it would be better to call `VERSION_RE.match` only within the `[requires]` section, ie just after the `elif section_name.endswith('require'):`, instead of on every line in every section.

### Used the version from the test as target-env to GLSLangValidator

We should do the same thing when calling spirv-as in assemble_stage.

If we do the above suggestion to make vr_requirements have a default value for the version, then we can get rid of `TARGET_ENV` and unconditonally use the value returned from `vr_requirements_get_version`.

hikiko:
ok, thanks for the feedback, I am working on it

bpeel:
Thanks for updating the branch.

I have a few more comments:

- The precompile script crashes if you it on a script with comments before the first section. For an example try `geometry.shader_test`. Maybe we should just add `section_name=None` add the top of `convert_stream`.
- The precompile script now removes all of the lines from the require section apart from the version line. Maybe it would be better to move the check for the version regexp to the bottom of the “else” section so that it always still calls `fout.write(line)` even in the requires section.
- It might be nice to tidy up the patches. In general it’s not ideal to have a branch that starts do things one way in early patches and then changes it later to do it a different way in later patches. We could just squash the later changes into the earlier patches to make a coherent set of changes.

You might want to try running the `test-build.sh` script to do some minimal testing when developing.

bpeel:
Thanks again for updating the branch.

I just have one final comment. It looks like the check for the version regexp in the precompile script was still only being used on section header lines so it wasn’t being picked up correctly. I’ve pushed a fixup commit onto your branch. If you think that is ok, feel free to squash it and merge your branch into master.

PR #64, opened 2019-05-10
Added support for long options

hikiko:
With this branch we could support both long and short options using an MIT licensed getopt implementation. We can use unique negative numbers to denote options without a short option.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -91,3 +91,31 @@ vr-list.h
  * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE
  * OF THIS SOFTWARE.
  */
+
+The following files taken from musl have the license below:
+
+getopt.h
+getopt.c
+
+/*
+ * Copyright 2005-2012 Rich Felker, et al.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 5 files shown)


bpeel (commented):
This looks good to me, thanks. I just have a couple of comments:

- Is there a good reason to put getopt file into a separate static library? Could we not just drop it directly into `src/getopt.c` and name it in VKRUNNERBIN_SOURCE_FILES?
- In `opt_help()`, it would be nice to make the format arg for the long options something like `%-7s` so that the columns line up properly.

hikiko:
> Is there a good reason to put getopt file into a separate static library? Could we not just drop it directly into src/getopt.c and name it in VKRUNNERBIN_SOURCE_FILES?

No :) I just wasn't sure if I should put the "borrowed" files together with the others, I will move them.

I've found another issue as well, long options that don't have short options are ignored the way I've written it, I will fix that too.

Also I am writing here the two other things we discussed on IRC so that we have all the needed fixes together:
1. I will edit the commit file to mention that this fixes issue 63
2. Add a license in the getopt.c (if the author gives permission to do so).

Thank you!

bpeel:
Thanks for making the changes. It looks good to me. Feel free to merge it.

PR #65, opened 2019-05-19
Allow selecting the Vulkan device

hikiko:
Users can select the Vulkan device using the option --device-id like follows:

`./vkrunner --device-id <device number>`
or
`./vkrunner --device-id=<device number>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,10 +416,11 @@ ubo 0 subdata int 0 \
     Runs the shader test script SCRIPT
     
     Options:
-      -h            Show this help message
-      -i IMG        Write the final rendering to IMG as a PPM image
-      -d            Show the SPIR-V disassembly
-      -D TOK=REPL   Replace occurences of TOK with REPL in the scripts
+      -h                Show this help message
+      -i IMG            Write the final rendering to IMG as a PPM image
+      -d                Show the SPIR-V disassembly
+      -D TOK=REPL       Replace occurences of TOK with REPL in the scripts
+      --device-id DEVID Select the Vulkan device
 
 ## Precompiling shaders
 
--- a/src/main.c
+++ b/src/main.c
@@ -32,6 +32,7 @@
 #include <errno.h>
 #include <stdint.h>
 #include <math.h>
+#include <limits.h>
 
 #include <vkrunner/vkrunner.h>
 
@@ -168,6 +169,22 @@ opt_quiet(struct main_data *data,
         return true;
 }
 
+static bool
+opt_device_id(struct main_data *data,
+              const char *arg)
+{
+        char *endp;
+        long id = strtol(arg, &endp, 0) - 1;
+        if (id < 0 || *endp) {
+                fprintf(stderr,
+                        "Invalid device id.\n");
+                return false;
+        }
+
+        vr_config_set_device_id(data->config, id);
+        return true;
+}
+
 /* Use unique negative numbers to denote options without a short option.
  * Don't choose -1, it's used in the getopt_long error check.
  */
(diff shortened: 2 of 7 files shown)


hikiko:
I forgot to write that it fixes issue 62 in the commit message and sent a second version

bpeel:
Thanks for the patches. I have some feedback below:

I think it would be better to add a function to `vr_config` to configure the device id rather than adding a parameter to `vr_executor_execute`. That way users of the API don’t need to think about it if they don’t want to specify the device and we won’t break any (hypothetical) existing applications using it. We can make `opt_device_id` directly call the new `vr_config_set_device_id` instead of having a field in `struct main_data` to set it.

If we do that maybe it doesn’t make sense to have the device ID on the `vr_requirements` any more because `find_physical_device` could just directly get the ID from the config. That way `vr_requirements_check` doesn’t need to duplicate the filtering by device ID either so it will be a bit cleaner. (I know I originally said to add it to the requirements, sorry about that).

Isn’t it missing something to make the first device ID number 1 in order to match CTS as described in the original issue?

Instead of calling `isprint()`, would it be better to just compare with >=0? Apparently the result of `isprint` is undefined if the value is not in the range 0 to 255 or EOF.

In the test-build.sh script, instead of removing the `set -u`, I think it would be better to just initialise `device_id` to an empty string.

hikiko:
The problem with -u was that it would return an error when the $1 or $2 are unset. According to the set documentation it is used to:

> Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’ as an error when performing parameter expansion. An error message will be written to the standard error, and a non-interactive shell will exit. 

so if I keep it the line with the check: 
`if [ "$1" = --device-id ]; then`

and run the ./test-build.sh without parameters I'll take the error message:
$1: unbound variable

Do we really want this back?

hikiko:
Thank you for the review, I've sent a new version, I only haven't done the -u change yet because I wasn't sure if we really need to remove it, I will fix it in a follow up branch if it's necessary.

bpeel:
> The problem with -u was that it would return an error when the $1 or $2 are unset.

Right, I see what you mean. However, I’d quite like to keep the `-u` if we can because it helps detect typos. Maybe we could change it to something like this?

    if [ $# -gt 0 ] && [ "$1" = "--device-id" ]; then

Or we could use a fancy variable expansion like this to make it evaluate to an empty string when $1 is unset:

    if [ "${1:-}" = "--device-id" ]; then

Thanks for updating the branch, I will take a look shortly.

bpeel:
Here is the remaining feedback for the v2.

The test for whether there is a short option is inversed and it’s currently removing all of the short options from the help.

I think we don’t need to add `vr_config_get_device_id`. There aren’t any other getters in the public API for the config and the less API we expose the better.

It looks like in `find_physical_id` if you use a device ID that is greater than the number of devices it would read off the end of the array? Maybe we should try to guard against that?

As far as I can tell it’s still missing something to make the IDs start from 1.

hikiko:
Thanks for the review I've just pushed the corrections

hikiko:
Hmm, sorry this only implements the changes of the last feedback, I am going to look at possible ways to put back the -u in the shell script and update it.

bpeel:
As well as adding back the `-u`, perhaps you could also look again at limiting the device ID to the range of devices. The are still two problems as far as I can see:

* If you specify an ID that is >= the number of devices it will look at all of the devices as if you didn’t specify an option instead of making the test skip.

* I think it’s impossible to select a device other than the first one. For example, if there are 6 devices and you set device_id to 2, then first_device will be set to 1 and count will also be set to 1. 1 is not less than 1 so the for loop will exit immediately.

bpeel:
Thanks for updating the branch.

Unfortunately it looks like `test-build.sh` now doesn’t work if you don’t specify the `--device-id` option because it references `$device_id` without first setting it. Perhaps you could initialise it to the empty string like I suggested in the first comment. The error message for when `--device-id` is given without an argument no longer works either because in that case `$2` won’t be set. Perhaps you could change it to `"${2:-}"`. I think it would be helpful to run the script before posting a new version of the branch to verify that it works.

The rest of the patches look good. Feel free to push those two patches in the meantime if you want close the issue before fixing the `test-build.sh` script as that is probably less important.

hikiko:
Merged, after @bpeel confirmed the final PR looks fine.

PR #67, opened 2019-09-30
Report 'skip' on instance creation failure.

kaydenl:
This causes vkrunner tests to report 'skip' rather than 'fail' when
run on a platform which doesn't support Vulkan and has no drivers.
The tests aren't running and failing - they're simply not executing.

For example, a CI system might deploy piglit and vkrunner on all
systems, but only some machines support Vulkan.  Similarly, Piglit
builds all OpenGL tests, but if the GPU doesn't support the required
API, then tests report 'skip'.

### DIFF
--- a/vkrunner/vr-context.c
+++ b/vkrunner/vr-context.c
@@ -262,7 +262,7 @@ init_vk_device(struct vr_context *context,
         if (res != VK_SUCCESS) {
                 vr_error_message(context->config,
                                  "Failed to create VkInstance");
-                return VR_RESULT_FAIL;
+                return VR_RESULT_SKIP;
         }
 
         vr_vk_init_instance(vkfn, get_instance_proc, context);


bpeel:
Thanks for the patch.

I’d sort of expect that if there are no drivers then the loader would still allow the instance to be created but it would just return a count of zero from `vkEnumeratePhysicalDevices`. However I can see that [that is not the case](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/loader.c#L6344).

Just in case of the unlikely event that creating the instance fails for some other reason, would it be worth limiting the SKIP case to when `res==VK_ERROR_INCOMPATIBLE_DRIVER`?

bpeel:
I’ve made a [v2 of the MR](https://github.com/Igalia/vkrunner/pull/69) with the change suggested above.

PR #68, opened 2020-02-11
Add support for descriptor set (ubo/ssbo) arrays

infapi00:


infapi00:
Just found a regression with this changes. Now tests using two simple ubos are failing. Will take a look to that problem.

infapi00:
> Just found a regression with this changes. Now tests using two simple ubos are failing. Will take a look to that problem.

I have just pushed a new version of the branch. The fix for this regression was added as a new commit, in order to explain it in detail, and to make easier the review. But probably it would make sense to squash both commits.

samuelig:
Repository moved to https://gitlab.freedesktop.org/mesa/vkrunner. Please re-submit this pull request there.

PR #70, opened 2020-07-08
update enum table to be compatible with Vulkan 1.2.141 headers

janesma:
Some enums were removed from the Vulkan headers, which generated
vkrunner build errors.


janesma:
@bpeel due to some unfortunate backward-incompatible changes made to vulkan headers, I had to re-run make-enums.py.  @djdeath suggested that vkrunner should just check in the version of vulkan headers for the API that it needs.  I could probably update your build support to do this if you think that is a good idea.

bpeel:
Thanks for the MR. However, doesn’t this remove all of the enums, not just the missing ones? Maybe `make-enums.py` needs updating if the format of the vulkan headers has also changed.

There is already [a branch](https://github.com/Igalia/vkrunner/commits/include-vulkan-header) which updates the build system to include the Vulkan headers. I made it to make it easier to integrate VkRunner into the CTS. I’m not really sure whether it’s a good idea, so I’m happy to hear any arguments for or against.

bpeel:
I just went ahead and merged a [modified version](https://github.com/Igalia/vkrunner/commit/f688a7498353c147b12f54ba523725bd108226a6) of the patch that only removes the appropriate enums. I also [updated](https://github.com/Igalia/vkrunner/commit/a513916ca2df5b2129553ead1d371fd8bb3b2296) the comment for how to run `make-enums.py` so that it works properly now. Hope that’s ok.

If we want to continue the discussion about importing the Vulkan header into the repo maybe we could open another issue.

PR #71, opened 2020-12-21
CmakeLists: fix install libdir bindir

samuelig:
Previously it was fixed to <path>/lib but some distros install 64-bit
libraries in <path>/lib64.

Fixed bindir as well.

Signed-off-by: Samuel Iglesias Gonsálvez <siglesias@igalia.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.2)
+cmake_minimum_required(VERSION 3.0.2)
 
 include(TestBigEndian)
 include(CheckLibraryExists)
@@ -7,7 +7,7 @@ include(CheckCCompilerFlag)
 include(FindPkgConfig)
 include(CheckIncludeFile)
 
-project(VkRunner)
+project(VkRunner LANGUAGES C)
 set(VKRUNNER_VERSION 0.1)
 
 TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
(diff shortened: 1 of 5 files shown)


bpeel:
Thanks for the patch.

It looks like it might need a bit more tweaking to work. When I try installing a build it ends up trying to install to `/pkg-config`, I guess because `$CMAKE_INSTALL_LIBDIR` isn’t defined. It looks like you need to do this in order to use it:

    include(GNUInstallDirs)

VkRunner is currently set to a minimum version of 2.82 of cmake, but that module seems to have been added in 3.0. I think we could safely changed the requirement to 3.0 because VK-GL-CTS also requires that now.

Next we seem to get a warning about not defining a language if we do that so I think we have to add this to the `project` define:

    project(VkRunner LANGUAGES C)

I think we’d have to fix the .pc file too because it’s currently hardcoded to `lib`.

samuelig:
Thanks for the review! I'll update the patch.

samuelig:
> Thanks for the patch.
> 
> It looks like it might need a bit more tweaking to work. When I try installing a build it ends up trying to install to `/pkg-config`, I guess because `$CMAKE_INSTALL_LIBDIR` isn’t defined. It looks like you need to do this in order to use it:
> 
> ```
> include(GNUInstallDirs)
> ```
> 
> VkRunner is currently set to a minimum version of 2.82 of cmake, but that module seems to have been added in 3.0. I think we could safely changed the requirement to 3.0 because VK-GL-CTS also requires that now.
> 

It was added in 2.8.5. However, to keep consistency to VK-GL-CTS I will updated it to 3.0.

> Next we seem to get a warning about not defining a language if we do that so I think we have to add this to the `project` define:
> 
> ```
> project(VkRunner LANGUAGES C)
> ```
> 
> I think we’d have to fix the .pc file too because it’s currently hardcoded to `lib`.

bpeel:
Thanks for the updates. It looks good to me.

I added an extra patch to your branch to fix the `test-build.sh` script. If you think that looks ok please push it along with your patches.

PR #72, opened 2023-01-11
box: Don’t apply the major axis setting to vector types

bpeel:
The major axis setting should only affect matrix types. Previously if a test script changed the layout to row-major in order to set up some matrix values and then also set up some vec values then the vec values would be stored incorrectly.

I don’t have commit access to the repo so if you think this patch is ok please merge it. Thanks!

I noticed this while working on a [port to rust](https://github.com/bpeel/vkrunner/commits/rust).


samuelig (approved):

samuelig:
Thanks Neil!

PR #73, opened 2023-03-02
Rewrite in Rust

bpeel:
This pull request replaces all of the C code with Rust, as mentioned in [this message](https://lists.freedesktop.org/archives/mesa-dev/2023-February/225926.html) to the mesa-dev mailing list.

The Rust port should be a drop-in replacement for the C version so everything should continue to work as before. The static library with the API is also still generated so it should still work, although presumably nothing is using that since CTS dropped VkRunner. I’ve tested building on Android and Windows (well, with MinGW+Wine) and that should still work too.

This PR switches the build system from CMake to Meson. I think the [Jenkins CI](https://gitlab.freedesktop.org/Mesa_CI/mesa_jenkins/-/blob/master/vkrunner/build.py) will try to automatically get the latest version from git and build with CMake. In order to try and avoid breaking that, I’ve added an extra patch on top of the branch to add some stub CMake files to divert to building with Meson. So, if the Jenkins CI has a rust compiler installed, there is a chance that it won’t break if this PR is merged, but I haven’t got confirmation of that.

I haven’t tried to clean up the commits too much so there are a couple of patches where I’ve implemented something and then later changed how it works or fixed a mistake. Not sure if it’d be worth trying to squash some of them. However, I tried to make sure that it always builds so all of the commits should still work if anyone ends up bisecting in them.

The advantages of merging this PR are:

- Going forward if someone works on VkRunner then it’s much nicer to write in Rust, particularly for doing string manipulation and not having to worry so much about cleaning up resources on errors.
- I’ve added a lot of unit tests so the Rust version is potentially more robust.
- I’ve included the bindgen of the Vulkan headers in the git repo so you don’t have to point the build scripts to the vulkan header any more. The project now has no dependencies at all (apart from the Rust compiler+stdlib).
- I added a patch to make VkRunner pass `--quiet` to `glslangvalidator` and `spirv-as` so it no longer does the annoying thing of printing the temporary filenames when running.
- I updated the `test-build.sh` script to use the right environment variable for the Vulkan validation layers. This picked up a problem with specifying the `VkImageAspectFlags` when clearing the depth buffer which wasn’t being picked up before. I’ve fixed that too.

The disadvantages are:

- Fewer people know Rust.
- It might break Jenkins.


1ace:
Thanks for the RiiR!  I have no Rust experience beyond trivial exercises, so I won't try to review the actual code.

Just a note for the last commit adding a cmake wrapper: I understand your goal, but personally I don't think it's a good idea because it will just cause some things to limp for a bit, but the wrapper isn't perfect (it can't be) so this will lead to subtle issues that can be avoided if the build just breaks "cleanly" for everyone and things get converted right away.
/cc @janesma for your opinion on this since you wrote the Intel CI integration; I do realize that it might not be as trivial as it looks to have something like `if Path("meson.build").exists() MesonBuilder(...) else CMakeBuilder(...)` so perhaps this is better in the short term?

I also noticed a couple of `if rust.found()` which surprise me: in what case could that be false? I'm guessing they might be remnants of the transition and can be removed now

Edit: I somehow missed the corresponding `required : false` on the rust import, which is what I should've asked about 😅

bpeel:
Thanks for taking a look at the PR!

> Just a note for the last commit adding a cmake wrapper: I understand your goal, but personally I don't think it's a good idea because it will just cause some things to limp for a bit, but the wrapper isn't perfect (it can't be) so this will lead to subtle issues that can be avoided if the build just breaks "cleanly" for everyone and things get converted right away.

Yes, you’re right, the wrapper probably isn’t a good idea. There’s no rush to merge the PR so it’d be better to try and find a proper solution for the Intel CI.

> /cc @janesma for your opinion on this since you wrote the Intel CI integration; I do realize that it might not be as trivial as it looks to have something like `if Path("meson.build").exists() MesonBuilder(...) else CMakeBuilder(...)` so perhaps this is better in the short term?

That sounds great if that works.

> I also noticed a couple of `if rust.found()` which surprise me: in what case could that be false? I'm guessing they might be remnants of the transition and can be removed now

Yes, I was probably just being over conservative and it’d be better to remove them. For a while in the version of Meson that I had from my Fedora install the `rust.found()` call failed, but I think it was a bug because if I set `required` to `true` and then just used the rust import unconditionally then it worked fine. Fedora seems to have Meson 1.0.0 now, so I guess there’s no point worrying about that any more.

markjanes-intel:
Thanks for pointing this out to me.  I'll get the CI in line to handle either mechanism.  @bpeel it's great to hear from you again!

bpeel:
I tried the Rust port on my Raspberry Pi and I realised that it doesn’t compile on 32-bit systems :/ It turns out that keeping the generated bindgen sources in the git repo is a really bad idea but even just making them be generated at build time isn’t enough to fix it. I think the Vulkan headers define the handles as pointers to opaque types on 64-bit systems but on 32-bit systems they are 64-bit integers. This breaks when I try to use `std::ptr::null()` instead of `VK_NULL_HANDLE` because it can’t implicitly convert from a pointer to an integer type.

I will try to fix this, but for now I’ll mark the PR as a draft.

bpeel:
I’ve remove the CMake wrapper and made bindgen be part of the build process. The meson rust module is now required to build.

The build is now a little bit awkward because bindgen needs the full path to the `vulkan.h` header file and it also needs the include path so that headers included by `vulkan.h` can be found. I’ve made it so it tries to find the directory that `vulkan/vulkan.h` is in by extracting the `includedir` variable from the vulkan pkg-config file. Then it can use that for the include path and also get the full path to the header file by appending `vulkan/vulkan.h`. Alternatively the Vulkan header directory can be specified with the `-Dvulkan_header_path=<path>` meson option.

cmarcelo:
@bpeel are you still working on this?

> It turns out that keeping the generated bindgen sources in the git repo is a really bad idea but even just making them be generated at build time isn’t enough to fix it.

Why is making them be generated at build time not enough?  Or maybe, could we vendor the Vulkan headers the same way Mesa and other projects do?

Context I am: to make certain changes that depend on new Vulkan headers, it is not clear to me how to ensure that a recent enough header will be available.  Also, as the headers evolve and fixes things (rename stuff e.g. VariablePointer vs VariablePointers feature name), it gets harder to make sure the Python scripts will work with both old and new headers.

bpeel:
Hi @cmarcelo

> @bpeel are you still working on this?

I haven’t touched it since the last comment, but I am happy to put some time back into it if need be.

> Why is making them be generated at build time not enough?

I think I meant to explain why generating them at build time wasn’t enough in the sentences just after that quote. So I guess the only problem is the thing about `VK_NULL_HANDLE`.

I updated the PR after making those comments. You can see the explanation on the comment from the [16th of March](https://github.com/Igalia/vkrunner/pull/73#issuecomment-1472395513). So basically the code in the PR does now invoke bindgen during the build and the `VK_NULL_HANDLE` issue is fixed.

> Or maybe, could we vendor the Vulkan headers the same way Mesa and other projects do?

What do you mean by vendoring the Vulkan headers? Do you mean to include them directly in the repo? Yes, that could be a good idea, especially if it is what Mesa is doing.

> Context I am: to make certain changes that depend on new Vulkan headers, it is not clear to me how to ensure that a recent enough header will be available. Also, as the headers evolve and fixes things (rename stuff e.g. VariablePointer vs VariablePointers feature name), it gets harder to make sure the Python scripts will work with both old and new headers.

The Python scripts aren’t run as part of the build process, instead they are run manually and the results are checked into the repo. Bearing that in mind I don’t think it’d be necessary to make the scripts continue to work with the old headers. I think it’d be fine to just ask anyone that runs the scripts to make sure they have the latest headers.

cmarcelo:
> I think it’d be fine to just ask anyone that runs the scripts to make sure they have the latest headers.

Not only the time we run the scripts but also need them when building. Imagine new header adds a new enumeration value to one of the enums we track, wouldn't the build fail with an older header due to missing symbol?

cmarcelo:
Yes, by "vendoring" I meant including the necessary Vulkan headers as part of the repository.

bpeel:
> Not only the time we run the scripts but also need them when building. Imagine new header adds a new enumeration value to one of the enums we track, wouldn't the build fail with an older header due to missing symbol?

Right yeah, I guess that’s true now that bindgen is run as part of the build. So putting the headers in the repo could be a nice solution 👍

cmarcelo:
One last comment: all things being equal, using cargo instead of meson+rust would be friendlier to various editor environments that support Rust (this will be at least my case).

bpeel:
VkRunner has now moved to Freedesktop.org. I’ve recreated the PR over [there](https://gitlab.freedesktop.org/mesa/vkrunner/-/merge_requests/1), so I’ll close this PR.

PR #74, opened 2023-04-24
Use vr_strcasecmp for Windows compatibility

Icohedron:
The `vr-util.h` has a preprocessor definition for `vr_strcasecmp` that selects the correct case-insensitive string comparison function for the current operating system.
However, `vr-util.c` contains uses of `strcasecmp` instead of `vr_strcasecmp`, which causes compilation to fail on Windows.


bpeel:
Thanks for the PR. It looks good to me.

I don’t have write access to merge it myself.

